PR #2, opened 2020-11-10
publish IsOsgVoContainer attribute

edquist:


brianhlin (approved):
LGTM! @rynge?

Also, @edquist are we populating the Site/ResourceName in the probe based on the startd attrs? If so, we'll want to update the container so those can be specified at runtime!

rynge:
LGTM as well. We already let the user publish the site identifiers:

       -e GLIDEIN_Site="..." \
       -e GLIDEIN_ResourceName="..." \

brianhlin:
Oh, perfect!

PR #3, opened 2021-02-05
add build-container workflow (SOFTWARE-4250)

edquist:
@brianhlin 

Largely based on the docker-frontier-squid workflow.

I am not sure if `opensciencegrid/osgvo-docker-pilot` is the right name for the dockerhub repo.

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -1,0 +1,30 @@
+name: dispatched build-docker-image
+
+on:
+  push:
+    branches:
+      - master
+  repository_dispatch:
+    types:
+      - dispatch-build
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    if: github.repository == 'opensciencegrid/osgvo-docker-pilot'
+    steps:
+
+    - name: checkout osgvo-docker-pilot
+      uses: actions/checkout@v2
+
+    - name: make date tag
+      id: mkdatetag
+      run: echo "::set-output name=dtag::$(date +%Y%m%d-%H%M)"
+
+    - name: Build and push Docker images
+      uses: docker/build-push-action@v1
+      with:
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+        repository: opensciencegrid/osgvo-docker-pilot
+        tags: fresh,${{ steps.mkdatetag.outputs.dtag }}
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -9,22 +9,42 @@ on:
       - dispatch-build
 
 jobs:
-  build:
+  make-date-tag:
     runs-on: ubuntu-latest
-    if: github.repository == 'opensciencegrid/osgvo-docker-pilot'
+    outputs:
+      dtag: ${{ steps.mkdatetag.outputs.dtag }}
     steps:
-
-    - name: checkout osgvo-docker-pilot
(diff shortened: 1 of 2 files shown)


brianhlin:
Hi @edquist this will work but I was hoping for the following tags:

- `development`, `development-<timestamp>`
- `testing`, `testing-<timestamp>`
- `release`, `release-<timetsamp>`

Based on the our base `development`, `testing`, and `release` tags here: https://hub.docker.com/r/opensciencegrid/software-base/tags?page=1&ordering=last_updated

.github/workflows/build-container.yml
```diff
@@ -0,0 +1,50 @@
...
+        tag_list=$docker_repo:$REPO,$docker_repo:$REPO-$TIMESTAMP
+        echo "::set-output name=taglist::$tag_list"
+
+    - name: Build and push Docker images
+      uses: docker/build-push-action@v2.2.0
+      with:
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+        repository: opensciencegrid/osgvo-docker-pilot
+        push: true
```
edquist, line 47: is `push` required?  I copied this in from software-base

.github/workflows/build-container.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: checkout osgvo-docker-pilot
+      uses: actions/checkout@v2
+
+    - id: generate-tag-list
+      env:
+        REPO: ${{ matrix.repo }}
+        TIMESTAMP: ${{ needs.make-date-tag.outputs.dtag }}
+      run: |
+        docker_repo=opensciencegrid/osgvo-docker-pilot
+        tag_list=$docker_repo:$REPO,$docker_repo:$REPO-$TIMESTAMP
```
edquist, line 38: do we really need the docker repo name in the tag?  Previously the build-and-push-docker-images step worked specifying `repository` and the unqualified tag names

edquist:
@brianhlin - is this closer?

Dockerfile
```diff
@@ -1,9 +1,14 @@
 FROM opensciencegrid/osg-wn
```
brianhlin, line 1: I do want this image to eventually be based off of `software-base` and supervisord on some level, i.e. this is good if we base `osg-wn` on `software-base`

@matyasselmeci any thoughts on whether or not we should base `osg-wn` on `software-base`?

brianhlin (approved):
LGTM for the description in the ticket. There may be some additional work wrt `osg-wn` and `software-base`

Dockerfile
matyasselmeci: It would be really hard to base osg-wn on software-base because osg-wn contains a huge set of tags built basically on the cartesian product of `{3.3,3.4,3.5} * {devel,testing,release} * {el6,el7,el8}` (minus whichever of those combinations are impossible) whereas software-base is only a small subset of those. (Although I wouldn't mind reducing osg-wn to a small subset of those as well...)

Also supervisord is for daemons -- not sure what daemons would be in the wn-client except for maybe cron to run a cron job to update CA certs and VO data or something.

brianhlin: Boy, we certainly should get rid of those 3.3/3.4/el6 tags.

> Also supervisord is for daemons -- not sure what daemons would be in the wn-client except for maybe cron to run a cron job to update CA certs and VO data or something.

We could certainly override the supervisord command but that does feel a little silly I guess. Instead, I guess we should base the backfill container off of `software-base` and install the appropriate WN packages.

PR #4, opened 2021-02-16
Fixups to the Docker build/push GHA workflow

brianhlin:


.github/workflows/build-container.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         REPO: ${{ matrix.repo }}
         TIMESTAMP: ${{ needs.make-date-tag.outputs.dtag }}
       run: |
-        docker_repo=opensciencegrid/osgvo-docker-pilot
+        docker_repo=${GITHUB_REPOSITORY/opensciencegrid\/docker-/opensciencegrid/}
```
edquist, line 39: tbh the hard-coded name was intentional here since the github/dockerhub repo names to not follow the normal docker-XYZ pattern.  This recipe that strips the leading `docker-` component works, but in this case only because the github and dockerhub repo names are identical.

.github/workflows/build-container.yml
```diff
@@ -39,12 +39,18 @@ jobs:
...
-      uses: docker/build-push-action@v2.2.0
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+
+    - name: Log in to Docker Hub
+      uses: docker/login-action@v1
       with:
         username: ${{ secrets.DOCKER_USERNAME }}
         password: ${{ secrets.DOCKER_PASSWORD }}
-        repository: opensciencegrid/osgvo-docker-pilot
```
edquist, line 46: Ok i thought there was some reason we needed to specify `repository` explicitly in `build-push-action@v2.2.0`

edquist (approved):
if it fixes the problem, great!

.github/workflows/build-container.yml
brianhlin: Yup, that's fine by me and I've been thinking that we should rename this repo to `docker-backfll-container` or something like that such that the image becomes `opensciencegrid/backfill-container`

.github/workflows/build-container.yml
brianhlin: Nope, the repo has to be specified as part of each tag now

PR #5, opened 2021-02-23
Use software-base (SOFTWARE-4250)

brianhlin:
We have some niceties built into `opensciencegrid/software-base`. Depends on https://github.com/opensciencegrid/docker-software-base/pull/30


brianhlin:
Note that https://github.com/opensciencegrid/osgvo-docker-pilot/pull/5/commits/6bcfda416da9fdd915af1201f1603e17e65edf56 would leak the token into container logs so I can revert that until we implement https://opensciencegrid.atlassian.net/browse/SOFTWARE-4371

10-htcondor.conf
```diff
@@ -0,0 +1,4 @@
+[condor_master]
+command=/sbin/condor_master -f
```
edquist, line 2: Minor, but officially the path is `/usr/sbin/condor_master` in the rpm.  I know rhel makes them the same thing with symlinks but i have mixed ideological feelings about relying on that..  (`/bin/entrypoint.sh` is another story since it is put there explicitly in the Dockerfile.)

Dockerfile
```diff
@@ -8,25 +11,21 @@ RUN useradd osg \
...
+        singularity \
  && yum clean all \
  && mkdir -p /etc/condor/passwords.d /etc/condor/tokens.d \
  && curl -s -o /usr/sbin/osgvo-user-job-wrapper https://raw.githubusercontent.com/opensciencegrid/osg-flock/master/job-wrappers/user-job-wrapper.sh \
  && curl -s -o /usr/sbin/osgvo-node-advertise https://raw.githubusercontent.com/opensciencegrid/osg-flock/master/node-check/osgvo-node-advertise \
  && chmod 755 /usr/sbin/osgvo-user-job-wrapper /usr/sbin/osgvo-node-advertise
 
-COPY entrypoint.sh /bin/
+COPY 10-setup-htcondor.sh /etc/osg/image-init.d/
+COPY 10-cleanup-htcondor.sh /etc/osg/image-cleanup.d/
```
edquist, line 26: I was not familiar with the semantics of the image init / cleanup .d dirs, but that seems like a nice separation.

edquist (approved):
A few thoughts but no objections to how you have it.

10-htcondor.conf
brianhlin: Good call! I had copied this from elsewhere

10-cleanup-htcondor.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+rm -rf $(condor_config_val LOCAL_DIR)
```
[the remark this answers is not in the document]
brianhlin: Hopefully made slightly less scary with the latest commit to check if it's actually a dir first

PR #6, opened 2021-03-05
Throw away supervisord logs to support unpriv containers

brianhlin:
Purdue is seeing issues where the container can't write to `/var/log`. I'm asking them to test these changes

### DIFF
--- a/10-htcondor.conf
+++ b/10-htcondor.conf
@@ -2,6 +2,6 @@
 command=/usr/sbin/condor_master -f
 autorestart=true
 startsecs=20
-stdout_logfile=/dev/fd/1
+stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
 redirect_stderr=true
--- a/10-htcondor.conf
+++ b/10-htcondor.conf
@@ -1,4 +1,4 @@
-[condor_master]
+[program:condor_master]
 command=/usr/sbin/condor_master -f
 autorestart=true
 startsecs=20
--- a/10-htcondor.conf
+++ b/10-htcondor.conf
@@ -2,3 +2,6 @@
 command=/usr/sbin/condor_master -f
 autorestart=true
 startsecs=20
+stdout_logfile=/dev/fd/1
+stdout_logfile_maxbytes=0
+redirect_stderr=true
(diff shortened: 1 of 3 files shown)


goughes:
I wasn't able to grab this new image from dockerhub, maybe it takes some time to show up? I did patch a local copy of the repo with your changes and built a new image though. Here's a new error.

```
<snip>
+++ condor_config_val LOG
+++ condor_config_val LOG
++ tail -F /tmp/osgvo-pilot-16964/log/MasterLog /tmp/osgvo-pilot-16964/log/StartLog
tail: cannot open '/tmp/osgvo-pilot-16964/log/MasterLog' for reading: No such file or directory
tail: cannot open '/tmp/osgvo-pilot-16964/log/StartLog' for reading: No such file or directory
2021-03-05 21:05:03,525 CRIT could not write pidfile /supervisord.pid
```

10-htcondor.conf
```diff
@@ -2,3 +2,6 @@
 command=/usr/sbin/condor_master -f
 autorestart=true
 startsecs=20
+stdout_logfile=/dev/fd/1
```
matyasselmeci, line 5:
```suggestion
stdout_logfile=/dev/stdout
```

supervisord.conf
```diff
@@ -0,0 +1,4 @@
+[supervisord]
+nodaemon=true
+logfile=/dev/null
+logfile_maxybtes=0
```
matyasselmeci, line 4:
```suggestion
logfile_maxbytes=0
```

10-htcondor.conf
brianhlin: I'm not sure that this will actually provide anything useful since HTCondor dumps its own logs

matyasselmeci: I think the suggestion and the original line are equivalent but yeah I think the whole block doesn't really help because it would only get us the master log.

brianhlin:
@goughes I've added a few commits and pushed a new image to the same tag, could you try the following container?

```
docker pull opensciencegrid/osgvo-docker-pilot:gh6.unpriv-logging
```

matyasselmeci (approved):

PR #8, opened 2021-03-17
Use a fixed UID for the `osg` user and run as that user

matyasselmeci:


brianhlin:
@rynge we discussed doing this because the setup bails if we're running as root https://github.com/opensciencegrid/osgvo-docker-pilot/blob/master/10-setup-htcondor.sh#L5-L8

rynge:
I think having a user with a specific UID in there for setup is ok, but a
requirement is to be able to run this container as any non-root UID. For
example, I should be able to do:

docker run -it --rm --user 4123:1246  opensciencegrid/osgvo-docker-pilot

And have the container and HTCondor run under that UID/GID.

brianhlin:
I think we have to abandon this because of the Singularity use case

PR #9, opened 2021-03-22
Add cvmfsexec (SOFTWARE-4549)

matyasselmeci:
- Use cvmfsexec to mount CVMFS repos if needed
- Don't run crond (doesn't work as non-root)
- Create log files before tailing them to solve "<FILE> has been replaced witha remote file" error from tail

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ You can store the logs outside of the container by bind-mounting a directory to
 
 cvmfsexec requires the additional options `--device=/dev/fuse` and
 `--security-opt=seccomp=unconfined`.  Supporting Singularity jobs when using
-cvmfsexec requires privileged containers (`--privileged`).
+cvmfsexec requires privileged containers (`--privileged`).  Using privileged containers avoids the need for those two additional options and all the `--cap-add` options without adding much risk so it is recommended.
 
 The following example invocation will:
 -   Use a token for authentication
--- a/README.md
+++ b/README.md
@@ -130,7 +130,8 @@ to `/cvmfs-cache`.
 You can store the logs outside of the container by bind-mounting a directory to
 `/cvmfs-logs`.
 
-cvmfsexec requires the additional option `--device=/dev/fuse`.
+cvmfsexec requires the additional options `--device=/dev/fuse` and
+(for Docker) `--security-opt=seccomp=unconfined`.
 
 Here is an example invocation using a token for authentication, using cvmfsexec
 to mount the OASIS and singularity repos instead of bind-mounting `/cvmfs`,
@@ -143,6 +144,7 @@ docker run -it --rm --user osg \
        --cap-add=DAC_READ_SEARCH \
        --cap-add=SYS_ADMIN --cap-add=SYS_CHROOT --cap-add=SYS_PTRACE \
        --device=/dev/fuse \
+       --security-opt=seccomp=unconfined \
        -v /var/cache/cvmfsexec:/cvmfsexec-cache \
        -v /var/log/cvmfsexec:/cvmfsexec-logs \
        -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org
--- a/README.md
+++ b/README.md
@@ -135,13 +135,15 @@ You can store the logs outside of the container by bind-mounting a directory to
 `/cvmfs-logs`.
 
 cvmfsexec requires the additional options `--device=/dev/fuse` and
-(for Docker) `--security-opt=seccomp=unconfined`.  Supporting Singularity
-jobs when using cvmfsexec requires privileged containers (`--privileged`).
+`--security-opt=seccomp=unconfined`.  Supporting Singularity jobs when using
+cvmfsexec requires privileged containers (`--privileged`).
 
-Here is an example invocation using a token for authentication, using cvmfsexec
-to mount the OASIS and singularity repos instead of bind-mounting `/cvmfs`,
-sending the cache to `/var/cache/cvmfsexec` and the logs to
-`/var/log/cvmfsexec`, and adding support for Singularity jobs:
+The following example invocation will:
+-   Use a token for authentication
+-   Use cvmfsexec to mount the OASIS and Singularity CVMFS repos
+-   Use `/var/cache/cvmfsexec` on the host for the CVMFS cache
+-   Use `/var/log/cvmfsexec` on the host for the CVMFS logs
+-   Support Singularity jobs
 
 ```
 docker run -it --rm --user osg \
--- a/README.md
+++ b/README.md
@@ -20,16 +20,19 @@ In order to successfully start payload jobs:
 In addition, you will be able to run more OSG jobs if you provide CVMFS.  You can do this
 in two ways:
 
-1. Mount CVMFS on the host and bind-mount it into the container by adding `-v /cvmfs:/cvmfs`.
+1. Mount CVMFS on the host and bind-mount it into the container by adding `-v /cvmfs:/cvmfs:shared`.
    This is the preferred mechanism.
 2. Use cvmfsexec as described in [the cvmfsexec section below](#cvmfs-without-a-bind-mount-using-cvmfsexec).
+
+Note: Supporting Singularity jobs inside the container will require the capabilities
+DAC_OVERRIDE, DAC_READ_SEARCH, SETGID, SETUID, SYS_ADMIN, SYS_CHROOT, and SYS_PTRACE.
 
 Example invocation utilizing a token for authentication:
 
 ```
 docker run -it --rm --user osg \
        --cap-add=DAC_OVERRIDE --cap-add=SETUID --cap-add=SETGID \
-       --cap-add=CAP_DAC_READ_SEARCH \
+       --cap-add=DAC_READ_SEARCH \
        --cap-add=SYS_ADMIN --cap-add=SYS_CHROOT --cap-add=SYS_PTRACE \
        -v /cvmfs:/cvmfs:shared \
        -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org
@@ -104,8 +107,16 @@ you can do in one of two ways:
     `--security-opt seccomp=unconfined --security-opt systempaths=unconfined --device=/dev/fuse`
     to the `docker run` invocation.
 
-The second option will add only the minimum necessary privileges for cvmfsexec.
-You can add additional security to that option by also adding `--security-opt no-new-privileges`.
+The first option is simpler and does not require unprivileged user namespaces
+to be enabled, but adds more privileges to the container.
+Using `--privileged` also adds the capabilities listed below for running
+Singularity jobs.
+
+The second option will add only the minimum necessary privileges for cvmfsexec,
+but requires unprivileged user namespaces to be enabled.
+You can add additional security to the second option by also adding
+`--security-opt no-new-privileges`.
+
 If cvmfsexec does not have the required privileges, the container will fail.
 
 Note that cvmfsexec will not be run if CVMFS repos are already available in
@@ -135,22 +146,25 @@ to `/cvmfs-cache`.
 You can store the logs outside of the container by bind-mounting a directory to
 `/cvmfs-logs`.
 
+Note: Supporting Singularity jobs inside the container will require the capabilities
+DAC_OVERRIDE, DAC_READ_SEARCH, SETGID, SETUID, SYS_ADMIN, SYS_CHROOT, and SYS_PTRACE.
 
 Here is an example invocation using a token for authentication, using cvmfsexec
 to mount the OASIS repos instead of bind-mounting `/cvmfs`, sending the cache
-to `/var/cache/cvmfsexec` and the logs to `/var/log/cvmfsexec`:
+to `/var/cache/cvmfsexec` and the logs to `/var/log/cvmfsexec`,
+and adding support for Singularity jobs:
 
 ```
 docker run -it --rm --user osg \
        --cap-add=DAC_OVERRIDE --cap-add=SETUID --cap-add=SETGID \
-       --cap-add=CAP_DAC_READ_SEARCH \
+       --cap-add=DAC_READ_SEARCH \
        --cap-add=SYS_ADMIN --cap-add=SYS_CHROOT --cap-add=SYS_PTRACE \
        --security-opt seccomp=unconfined \
        --security-opt systempaths=unconfined \
        --device=/dev/fuse \
        -v /var/cache/cvmfsexec:/cvmfsexec-cache \
        -v /var/log/cvmfsexec:/cvmfsexec-logs \
-       -e TOKEN="..." \
+       -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org
        -e GLIDEIN_Site="..." \
        -e GLIDEIN_ResourceName="..." \
        -e GLIDEIN_Start_Extra="True" \
--- a/README.md
+++ b/README.md
@@ -131,7 +131,8 @@ You can store the logs outside of the container by bind-mounting a directory to
 `/cvmfs-logs`.
 
 cvmfsexec requires the additional options `--device=/dev/fuse` and
-(for Docker) `--security-opt=seccomp=unconfined`.
+(for Docker) `--security-opt=seccomp=unconfined`.  Supporting Singularity
+jobs when using cvmfsexec requires privileged containers (`--privileged`).
 
 Here is an example invocation using a token for authentication, using cvmfsexec
 to mount the OASIS and singularity repos instead of bind-mounting `/cvmfs`,
@@ -140,11 +141,7 @@ sending the cache to `/var/cache/cvmfsexec` and the logs to
 
 ```
 docker run -it --rm --user osg \
-       --cap-add=DAC_OVERRIDE --cap-add=SETUID --cap-add=SETGID \
-       --cap-add=DAC_READ_SEARCH \
-       --cap-add=SYS_ADMIN --cap-add=SYS_CHROOT --cap-add=SYS_PTRACE \
-       --device=/dev/fuse \
-       --security-opt=seccomp=unconfined \
+       --privileged
        -v /var/cache/cvmfsexec:/cvmfsexec-cache \
        -v /var/log/cvmfsexec:/cvmfsexec-logs \
        -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org \
--- a/README.md
+++ b/README.md
@@ -112,7 +112,8 @@ There are several environment variables you can set for cvmfsexec:
 
 -   `CVMFSEXEC_REPOS` - this is a comma-separated list of CVMFS repos to mount,
     if using cvmfsexec; leave this blank to disable cvmfsexec.
-    OSG jobs frequently use the OASIS repo (`oasis.opensciencegrid.org`).
+    OSG jobs frequently use the OASIS repo (`oasis.opensciencegrid.org`) and
+    the singularity repo (`singularity.opensciencegrid.org`).
 
 -   `CVMFS_HTTP_PROXY` - this sets the proxy to use for CVMFS; if left blank
     it will find the best one via WLCG Web Proxy Auto Discovery.
@@ -132,9 +133,9 @@ You can store the logs outside of the container by bind-mounting a directory to
 cvmfsexec requires the additional option `--device=/dev/fuse`.
 
 Here is an example invocation using a token for authentication, using cvmfsexec
-to mount the OASIS repos instead of bind-mounting `/cvmfs`, sending the cache
-to `/var/cache/cvmfsexec` and the logs to `/var/log/cvmfsexec`,
-and adding support for Singularity jobs:
+to mount the OASIS and singularity repos instead of bind-mounting `/cvmfs`,
+sending the cache to `/var/cache/cvmfsexec` and the logs to
+`/var/log/cvmfsexec`, and adding support for Singularity jobs:
 
 ```
 docker run -it --rm --user osg \
@@ -149,7 +150,7 @@ docker run -it --rm --user osg \
        -e GLIDEIN_ResourceName="..." \
        -e GLIDEIN_Start_Extra="True" \
        -e OSG_SQUID_LOCATION="..." \
-       -e CVMFSEXEC_REPOS=oasis.opensciencegrid.org \
+       -e CVMFSEXEC_REPOS="oasis.opensciencegrid.org singularity.opensciencegrid.org" \
        opensciencegrid/osgvo-docker-pilot:latest
 ```
 
--- a/README.md
+++ b/README.md
@@ -82,8 +82,8 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
 If you don't have CVMFS available on the host, the container can still make
 CVMFS available by using [cvmfsexec](https://github.com/cvmfs/cvmfsexec#readme).
 
-This will require a kernel version >= 3.10.0-1127 on an EL7-compatible host
-or >= 4.18 on an EL8-compatible host, or user namespaces enabled by hand --
+This will require a kernel version >= 3.10.0-1127 on an EL7-compatible host with user namespaces enabled
+or >= 4.18 on an EL8-compatible host --
 see the cvmfsexec README linked above for details.
 
 This will also require granting the container some additional privileges, which
@@ -95,7 +95,7 @@ you can do in one of two ways:
     `--security-opt seccomp=unconfined --security-opt systempaths=unconfined --device=/dev/fuse`
     to the `docker run` invocation.
 
-The second option will add only the minimum necessary privileges for cvmfsexec.
+The second option will add only the minimum necessary privileges for cvmfsexec.  You can add additional security to that option by also adding `--security-opt no-new-privileges`.
 
 Using cvmfsexec takes place in the entrypoint, which means it will still happen
 even if you specify a different command to run, such as `bash`.  You can bypass
@@ -105,15 +105,15 @@ clears the command.
 
 There are several environment variables you can set for cvmfsexec:
 
--   `NO_CVMFSEXEC` - set this to disable cvmfsexec.
+-   `NO_CVMFSEXEC` - set this to any non-empty value to disable cvmfsexec.
 
 -   `CVMFSEXEC_REPOS` - this is a comma-separated list of CVMFS repos to mount, if using cvmfsexec.
     Set this to blank to avoid using cvmfsexec, in which case you won't need to
     add the above permissions to your container.  The default is to mount the
     OASIS repo (oasis.opensciencegrid.org).
 
--   `CVMFS_HTTP_PROXY` - this sets the proxy to use for CVMFS; leave this blank
-    to find the best one via wlcg-lpad.
+-   `CVMFS_HTTP_PROXY` - this sets the proxy to use for CVMFS; if left blank
+    it will find the best one via WLCG Web Proxy Auto Discovery.
 
 -   `CVMFS_QUOTA_LIMIT` - the quota limit in MB for CVMFS; leave this blank to
     use the system default (4 GB)
@@ -153,4 +153,3 @@ cvmfsexec will not be used if:
 - /cvmfs is already available via bind-mount
 - `CVMFSEXEC_REPOS` is empty
 - `NO_CVMFSEXEC` is set
-
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ in two ways:
 Note: Supporting Singularity jobs inside the container will require the capabilities
 DAC_OVERRIDE, DAC_READ_SEARCH, SETGID, SETUID, SYS_ADMIN, SYS_CHROOT, and SYS_PTRACE.
 
-Example invocation utilizing a token for authentication:
+Example invocation utilizing a token for authentication and bind-mounting CVMFS:
 
 ```
 docker run -it --rm --user osg \
@@ -94,9 +94,10 @@ CVMFS available by using [cvmfsexec](https://github.com/cvmfs/cvmfsexec#readme).
 
 You will need to specify a list of CVMFS repos to mount in the environment
 variable `CVMFSEXEC_REPOS`.
-This will require a kernel version >= 3.10.0-1127 on an EL7-compatible host
-with user namespaces enabled or >= 4.18 on an EL8-compatible host --
-see the cvmfsexec README linked above for details.
+
+On EL7, you must have kernel version >= 3.10.0-1127 (run `rpm -q kernel` to check),
+and user namespaces enabled.  On EL8, you must have kernel version >= 4.18.
+See the cvmfsexec README linked above for details.
 
 Note that cvmfsexec will not be run if CVMFS repos are already available in
 `/cvmfs` via bind-mount.
--- a/README.md
+++ b/README.md
@@ -136,7 +136,9 @@ You can store the logs outside of the container by bind-mounting a directory to
 
 cvmfsexec requires the additional options `--device=/dev/fuse` and
 `--security-opt=seccomp=unconfined`.  Supporting Singularity jobs when using
-cvmfsexec requires privileged containers (`--privileged`).  Using privileged containers avoids the need for those two additional options and all the `--cap-add` options without adding much risk so it is recommended.
+cvmfsexec requires privileged containers (`--privileged`).  Using privileged
+containers avoids the need for those two additional options and all the
+`--cap-add` options without adding much risk, so it is recommended.
 
 The following example invocation will:
 -   Use a token for authentication
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ docker run -it --rm --user osg \
        --cap-add=DAC_READ_SEARCH \
        --cap-add=SYS_ADMIN --cap-add=SYS_CHROOT --cap-add=SYS_PTRACE \
        -v /cvmfs:/cvmfs:shared \
-       -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org
+       -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org \
        -e GLIDEIN_Site="..." \
        -e GLIDEIN_ResourceName="..." \
        -e GLIDEIN_Start_Extra="True" \
@@ -147,7 +147,7 @@ docker run -it --rm --user osg \
        --security-opt=seccomp=unconfined \
        -v /var/cache/cvmfsexec:/cvmfsexec-cache \
        -v /var/log/cvmfsexec:/cvmfsexec-logs \
-       -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org
+       -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org \
        -e GLIDEIN_Site="..." \
        -e GLIDEIN_ResourceName="..." \
        -e GLIDEIN_Start_Extra="True" \
@@ -155,4 +155,3 @@ docker run -it --rm --user osg \
        -e CVMFSEXEC_REPOS="oasis.opensciencegrid.org singularity.opensciencegrid.org" \
        opensciencegrid/osgvo-docker-pilot:latest
 ```
-
--- a/README.md
+++ b/README.md
@@ -140,6 +140,9 @@ There are several environment variables you can set for cvmfsexec:
 -   `CVMFS_QUOTA_LIMIT` - the quota limit in MB for CVMFS; leave this blank to
     use the system default (4 GB)
 
+You can add other CVMFS options by bind-mounting a config file over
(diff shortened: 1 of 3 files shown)


brianhlin (commented):
Why add this as `entrypoint.sh` instead of adding it a script to `/etc/osg/image-init.d`? @DrDaveD could you take a look at the `cvmfsexec` bits?

edquist:
Zing.

matyasselmeci:
Because I want cvmfs to be mounted regardless of the command being run.

matyasselmeci:
In this case, entrypoint.sh is used as the docker ENTRYPOINT, not the CMD; cvmfsexec is actually a wrapper around running some other command. I don't think it will work if I put it into /etc/osg/image-init.d since it does not affect the currently running process.

brianhlin:
> In this case, entrypoint.sh is used as the docker ENTRYPOINT, not the CMD; cvmfsexec is actually a wrapper around running some other command. I don't think it will work if I put it into /etc/osg/image-init.d since it does not affect the currently running process.

Oh, so the idea is that users of this container can pass runtime args to `cvmfsexec`? Could you add some comments to the `Dockerfile` to explain why we're doing it this way?

matyasselmeci:
I'll update the readme since it looks like it is confusing.

You're not passing runtime arguments to cvmfsexec (except via the new environment variables CVMFSEXEC_DIST and CVMFSEXEC_REPOS. Instead, you are passing runtime arguments to whatever command you're running.

For example,
`docker run ... opensciencegrid/osgvo-docker-pilot`
will run supervisord with CVMFS mounted.
`docker run ... opensciencegrid/osgvo-docker-pilot bash`
will run bash with CVMFS mounted.

on the other hand,
`docker run ... --entrypoint /bin/bash openscienceegrid/osgvo-docker-pilot`
will run bash without going through entrypoint.sh, so CVMFS won't be mounted.

brianhlin (changes requested):
This looks reasonable to me but I'd like some comments explaining why we set up `entrypoint.sh` like we do and documentation for the order in which this image tries to support CVMFS

entrypoint.sh
```diff
@@ -0,0 +1,25 @@
...
+    cvmfsexec_root=/tmp/cvmfsexec-$(id -u)
+    if [[ ! -d $cvmfsexec_root ]]; then
+        echo "No cvmfsexec dir found for this user at $cvmfsexec_root; creating one"
+        cp -rp /cvmfsexec $cvmfsexec_root || fail "Couldn't create $cvmfsexec_root"
+        echo "Fetching $CVMFSEXEC_DIST CVMFS config"
+        $cvmfsexec_root/makedist $CVMFSEXEC_DIST || fail "Couldn't fetch CVMFS config"
+    fi
+    $cvmfsexec_root/cvmfsexec -- /bin/true || \
+        fail "cvmfsexec smoke test failed.  You may not have the permissions to run cvmfsexec; see https://github.com/cvmfs/cvmfsexec#README for details"
+    exec $cvmfsexec_root/cvmfsexec $CVMFSEXEC_REPOS -- "$@"
```
DrDaveD, line 22: There should also separately be some logic in the pilot, such that if it knows that a repository needs to be mounted, it isn't mounted, and $CVMFSMOUNT is not empty, then it should invoke $CVMFSMOUNT to mount it.  Also the pilot needs to invoke `exec {CVMFSEXEC_CMDFD}>&-` if $CVMFSEXEC_CMDFD is set, to prevent giving the job access to the cvmfsexec command.

DrDaveD (changes requested):

DrDaveD:
I think there also should be a CVMFS_HTTP_PROXY configuration variable available which, when set, allows the container user to specify a proxy or list of proxies to override the default in /cvmfsexec/dist/etc/cvmfs/default.local.  Also  CVMFS_QUOTA_LIMIT, if it is set, so people can easily give cvmfs more than the default 4GB of cache if they have the space.  Both of these options are called out in the [README](https://github.com/cvmfs/cvmfsexec#making-the-cvmfs-distribution).

entrypoint.sh
DrDaveD: @mambelli mmascher The GlideinWMS cvmfsexec support should check for cvmfsexec to already be running.   If it is, the pilot should skip invoking it, but still do the other things it needs to do to play well with cvmfsexec.

DrDaveD: (@mmascher I accidentally left out the '@' for you on the previous comment.  I edited it in but I'm not sure if github would send a notice in response to an edit, so I add this extra comment.)

entrypoint.sh
```diff
@@ -0,0 +1,25 @@
...
+cvmfs_mounts=(/cvmfs/*)
+if [[ -z $cvmfs_mounts && -n $CVMFSEXEC_REPOS && -n $CVMFSEXEC_DIST ]]; then
+    echo "No cvmfs mounts; using cvmfsexec to mount $CVMFSEXEC_REPOS"
+    # We need our own copy of cvmfsexec for permissions reasons.
+    cvmfsexec_root=/tmp/cvmfsexec-$(id -u)
+    if [[ ! -d $cvmfsexec_root ]]; then
+        echo "No cvmfsexec dir found for this user at $cvmfsexec_root; creating one"
+        cp -rp /cvmfsexec $cvmfsexec_root || fail "Couldn't create $cvmfsexec_root"
+        echo "Fetching $CVMFSEXEC_DIST CVMFS config"
+        $cvmfsexec_root/makedist $CVMFSEXEC_DIST || fail "Couldn't fetch CVMFS config"
```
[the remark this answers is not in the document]
matyasselmeci: I tried that, but the intent is for this image to be runnable as any UID, and I got problems when running as a different UID than the one that did the makedist.

entrypoint.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+fail () {
+    echo "$@" >&2
+    exit 1
+}
+
+shopt -s nullglob
+cvmfs_mounts=(/cvmfs/*)
```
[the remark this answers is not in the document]
matyasselmeci: Sure, sounds good.

entrypoint.sh
```diff
@@ -0,0 +1,25 @@
...
+if [[ -z $cvmfs_mounts && -n $CVMFSEXEC_REPOS && -n $CVMFSEXEC_DIST ]]; then
+    echo "No cvmfs mounts; using cvmfsexec to mount $CVMFSEXEC_REPOS"
+    # We need our own copy of cvmfsexec for permissions reasons.
+    cvmfsexec_root=/tmp/cvmfsexec-$(id -u)
+    if [[ ! -d $cvmfsexec_root ]]; then
+        echo "No cvmfsexec dir found for this user at $cvmfsexec_root; creating one"
+        cp -rp /cvmfsexec $cvmfsexec_root || fail "Couldn't create $cvmfsexec_root"
```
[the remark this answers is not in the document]
matyasselmeci: The copy in /tmp is owned by the user that's running the container, which may not be the user that created the container.

entrypoint.sh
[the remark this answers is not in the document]
DrDaveD: If you do a makedist -o cvmfsexec (in a different directory or different name so as not overwrite cvmfsexec script) during the container build, then when cvmfsexec is executed it at run time it should self-extract as the user that is executing.  Alternatively maybe you could do a chown -R on the cvmfsexec directory.

entrypoint.sh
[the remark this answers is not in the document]
DrDaveD: Maybe chown -R would be lighter weight.

entrypoint.sh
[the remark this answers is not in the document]
DrDaveD: On second thought I think using makedist -o isn't a good idea, because there ought to be a way for people to bind mount a new default.local, cache directory, and log directory.  So it's better to have it be unpacked.  However I don't understand why the cp -r to /tmp didn't make everything writable.

entrypoint.sh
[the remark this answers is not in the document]
DrDaveD: Except that probably this script isn't running as fake root so chown wouldn't work.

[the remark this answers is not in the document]
matyasselmeci: Exactly. I create a "template" in /cvmfsexec which each user (as themselves) copies to create their own cvmfsexec directory into which they have write permissions.

I wonder if it would be a good idea to run makedist at image creation, tar up the results and, at startup time have, have the user untar the results to create their own copy...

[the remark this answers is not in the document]
DrDaveD: That's better than downloading at start time, but has the same issues as makedist -o I brought up in the [related thread on this ticket](https://github.com/opensciencegrid/osgvo-docker-pilot/pull/9#discussion_r599981246).  Follow up there.

entrypoint.sh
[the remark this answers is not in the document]
matyasselmeci: makedist creates dist/var/lib/cvmfs with 0700 permissions so cp -r can't read it.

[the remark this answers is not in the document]
DrDaveD: That comes from the cvmfs rpm so I guess you'll have to either chmod it, rmdir it and recreate it on the destination, or use a copy mechanism where you can exclude it.

Copying all the files to a new location complicates allowing people to bind mount in cache and log directories too.  I think it would probably be best to skip both dist/var/lib/cvmfs and dist/var/log and manipulate them manually.  If somebody has bind mounted either one at a documented location, create a symlink pointing back to them, otherwise I guess leave them in /tmp.  By looking at `mount` inside of a docker container it looks like the whole root filesystem is an overlay with the layers in /var/lib/docker/overlay2 so I guess that's OK as a default, but /var has to be big enough.  I think it would make sense to support bind-mounting at the paths /var/lib/cvmfs and /var/log/cvmfs, and also /etc/cvmfs/default.local.

README.md
```diff
@@ -76,4 +76,80 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
 ```
 
 
+## CVMFS Without a Bind-Mount Using cvmfsexec
+
+If you don't have CVMFS available on the host, the container can still make
+CVMFS available by using [cvmfsexec](https://github.com/cvmfs/cvmfsexec#readme).
+
+This will require a kernel version >= 3.10.0-1127 on an EL7-compatible host
+or >= 4.18 on an EL8-compatible host, or user namespaces enabled by hand --
```
DrDaveD, line 85:
```suggestion
This will require a kernel version >= 3.10.0-1127 on an EL7-compatible host with user namespaces enabled
or >= 4.18 on an EL8-compatible host --
```

README.md
```diff
@@ -76,4 +76,80 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+This will also require granting the container some additional privileges, which
+you can do in one of two ways:
+
+1.  Add `--privileged` to the `docker run` invocation.
+
+2.  Add
+    `--security-opt seccomp=unconfined --security-opt systempaths=unconfined --device=/dev/fuse`
+    to the `docker run` invocation.
+
+The second option will add only the minimum necessary privileges for cvmfsexec.
```
DrDaveD, line 97:
```suggestion
The second option will add only the minimum necessary privileges for cvmfsexec.  You can add additional security to that option by also adding `--security-opt no-new-privileges`.
```

README.md
```diff
@@ -76,4 +76,80 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+
+Using cvmfsexec takes place in the entrypoint, which means it will still happen
+even if you specify a different command to run, such as `bash`.  You can bypass
+the entrypoint by passing `--entrypoint <cmd>` where `<cmd>` is some different
+command to run, e.g. `--entrypoint bash`.  Setting the entrypoint this way
+clears the command.
+
+There are several environment variables you can set for cvmfsexec:
+
+-   `NO_CVMFSEXEC` - set this to disable cvmfsexec.
```
DrDaveD, line 107:
```suggestion
-   `NO_CVMFSEXEC` - set this to any non-empty value to disable cvmfsexec.
```

README.md
```diff
@@ -76,4 +76,80 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+
+-   `CVMFSEXEC_REPOS` - this is a comma-separated list of CVMFS repos to mount, if using cvmfsexec.
+    Set this to blank to avoid using cvmfsexec, in which case you won't need to
+    add the above permissions to your container.  The default is to mount the
+    OASIS repo (oasis.opensciencegrid.org).
+
+-   `CVMFS_HTTP_PROXY` - this sets the proxy to use for CVMFS; leave this blank
+    to find the best one via wlcg-lpad.
```
DrDaveD, line 115:
```suggestion
-   `CVMFS_HTTP_PROXY` - this sets the proxy to use for CVMFS; if left blank
    it will find the best one via WLCG Web Proxy Auto Discovery.
```

README.md
```diff
@@ -76,4 +76,80 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+
+Here is an example invocation using a token for authentication, using cvmfsexec
+instead of bind-mounting `/cvmfs`, sending the cache to `/var/cache/cvmfsexec`
+and the logs to `/var/log/cvmfsexec`:
+
+```
+docker run -it --rm --user osg \
+       --cap-add=DAC_OVERRIDE --cap-add=SETUID --cap-add=SETGID \
+       --cap-add=CAP_DAC_READ_SEARCH \
+       --cap-add=SYS_ADMIN --cap-add=SYS_CHROOT --cap-add=SYS_PTRACE \
```
DrDaveD, line 135: These capabilities are almost as bad as --privileged, especially SYS_ADMIN.  They should be unnecessary if you are using the options for low privilege.

README.md
```diff
@@ -76,4 +76,80 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+       -e OSG_SQUID_LOCATION="..." \
+       opensciencegrid/osgvo-docker-pilot:latest
+```
+
+
+If cvmfsexec does not have the required privileges, the container will fail.
+cvmfsexec will not be used if:
+- /cvmfs is already available via bind-mount
+- `CVMFSEXEC_REPOS` is empty
+- `NO_CVMFSEXEC` is set
```
DrDaveD, line 154: Make clear this is an "or" list and not an "and" list.  Why have both NO_CVMFSEXEC and CVMFSEXEC_REPOS empty to disable?  Isn't NO_CVMFSEXEC redundant?

README.md
```diff
@@ -76,4 +76,80 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+
+-   `CVMFS_QUOTA_LIMIT` - the quota limit in MB for CVMFS; leave this blank to
+    use the system default (4 GB)
+
+
+You can store the cache outside of the container by bind-mounting a directory
+to `/cvmfs-cache`.
+You can store the logs outside of the container by bind-mounting a directory to
+`/cvmfs-logs`.
+
```
DrDaveD, line 125:
```suggestion
You can customize cvmfs further by bind-mounting a file on
`/cvmfsexec/dist/etc/cvmfs/default.local`.
```

DrDaveD: Oh but now I realize this won't work because of the tarball extraction.  You would have to choose another path.  How about `/cvmfsexec/default.local` and then look for that and copy it into the final path if present.

DrDaveD (changes requested):
I would just like to see support for bind mounting a user-defined default.local.  The rest looks good other than documentation.  I also suggest removing NO_CVMFSEXEC as unnecessary.

README.md
matyasselmeci: Sure, that would work. If you bind-mount `/cvmfsexec/default.local` but also set the environment variables such as `CVMFS_HTTP_PROXY`, which should win?

README.md
matyasselmeci: I copied these from the original example (which bind-mounted CVMFS). Perhaps @rynge would know why all of those are needed?

README.md
DrDaveD: I think you could do it whichever way is more convenient.  I would guess given your current implementation that the variables would win.

README.md
DrDaveD: It was an attempt to do something less than `--privileged` but they didn't know about the completely unprivileged option.

matyasselmeci:
NO_CVMFSEXEC is technically redundant, but as I was putting together the readme, I realized that disabling a feature as a side effect of emptying a list sounded confusing, so I wanted to make a single, obvious option for "I'm not ready for this yet, how do I turn it off so my container starts?"

This sounds a little weird, but I think another way to make it less confusing (instead of NO_CVMFSEXEC) would be to keep current behavior but make CVMFSEXEC_REPOS be empty by default. So the instructions to the admin would be like

> If you want CVMFS, either
> 1. Mount CVMFS on the host and pass `-v /cvmfs:/cvmfs` or
> 2. Pass `-e CVMFSEXEC_REPOS=oasis.opensciencegrid.org --security-opt ...`

Does that make sense? @brianhlin , do you have an opinion on this?

bbockelm:
Is there any obvious way to mount "all" CVMFS repositories?

DrDaveD:
> Is there any obvious way to mount "all" CVMFS repositories?

One could download the list of repositories hosted by oasis-replica.opensciencegrid.org (although it wouldn't be robust, what if it is down, what's the backup) or extract the list of supported OSG repositories from topology, but it would very likely be time-prohibitive.  cvmfsexec supports mounting additional repositories inside if the command it runs, so the better thing to do is for the pilot to request additional mounts when needed.

matyasselmeci:
OK, updated. I got rid of NO_CVMFSEXEC and instead made CVMFSEXEC_REPOS be blank by default. This way, an admin can just update their image without changing their `docker run` command to keep current behavior.

@DrDaveD , can you take another look?

matyasselmeci:
also @brianhlin, are my readme changes good?

DrDaveD:
Ok on the removal of NO_CVMFSEXEC but I'd still like to see support for /cvmfsexec/default.local and removal of all those extraneous --cap-add options.

README.md
```diff
@@ -76,4 +83,78 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+see the cvmfsexec README linked above for details.
+
+This will also require granting the container some additional privileges, which
+you can do in one of two ways:
+
+1.  Add `--privileged` to the `docker run` invocation.
+
+2.  Add
+    `--security-opt seccomp=unconfined --security-opt systempaths=unconfined --device=/dev/fuse`
+    to the `docker run` invocation.
```
brianhlin, line 104: What's the use case for 1 vs 2?

brianhlin (commented):
Only one question about the README but otherwise it looks ok to me after a quick skim

README.md
DrDaveD: The first case is if system admins are comfortable with the risk and don't want to enable unprivileged user namespaces.  The second case is less risky but requires enabling unprivileged user namespaces.

matyasselmeci: 2 uses fewer privileges but it might not be possible to do in Kubernetes (seccomp requires the cluster admin to set something up, and I found absolutely nothing on systempaths).

brianhlin: @matyasselmeci please elaborate on 1 vs 2 using your + Dave's comments

DrDaveD:
You're not giving Kubernetes instructions at all, just docker, so I wouldn't mention that part unless you want to add an additional discussion on that.

Having said that, the equivalent of the systempaths=unconfined is apparently available in Kubernetes.  It is harder to enable, requiring setting a feature gate to enable [allowedProcMountTypes](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#allowedprocmounttypes).  I don't admin my own Kubernetes cluster so I haven't been able to try to get it to work myself.  One admin tried and failed, and Jeff said last week Tuesday he hoped to be ready to try it on Tiger by the time of the next kubernetes hackathon.

matyasselmeci:
OK, I added that `--privileged` was simpler and didn't require user namespaces, but gave more permissions to the container. I also explained what all the capabilities were used for (i.e. Singularity), and that `--privileged` implied them.

We can review which capabilities are needed and for what, but I would prefer to do that in a separate PR so we can keep the 'bind-mounted CVMFS' and 'CVMFS-via-cvmfsexec' command lines in sync.

I also added support for default.local.  Note that by default there is already stuff in `/cvmfsexec/dist/etc/cvmfs/default.local`:

```
CVMFS_HTTP_PROXY="auto;DIRECT"
CVMFS_PAC_URLS="http://grid-wpad/wpad.dat;http://wpad/wpad.dat;http://cernvm-wpad.fnal.gov/wpad.dat;http://cernvm-wpad.cern.ch/wpad.dat"
```
so anything the user mounts into `/cvmfsexec/default.local` will stomp over that.

README.md
```diff
@@ -16,14 +16,24 @@ In order to successfully start payload jobs:
...
    the pilot to only run certain jobs.
 
+In addition, you will be able to run more OSG jobs if you provide CVMFS.  You can do this
+in two ways:
+
+1. Mount CVMFS on the host and bind-mount it into the container by adding `-v /cvmfs:/cvmfs:shared`.
+   This is the preferred mechanism.
+2. Use cvmfsexec as described in [the cvmfsexec section below](#cvmfs-without-a-bind-mount-using-cvmfsexec).
+
+Note: Supporting Singularity jobs inside the container will require the capabilities
```
DrDaveD, line 26: No, this is still confused.  The same options needed for cvmfsexec work for singularity.  You either need to list the capabilities, use --privileged, or use the options for unprivileged.  I'm sorry I didn't make that clear; in fact I didn't realize that the discussion was distinct.

I suggest listing the 3 options here and not mention them under cvmfsexec.

DrDaveD (changes requested):

DrDaveD:
> so anything the user mounts into `/cvmfsexec/default.local` will stomp over that.

Yes that's the way that cvmfsexec expects it.   It only supplies those defaults if default.local hasn't been supplied.

README.md
matyasselmeci: OK, I ended up skipping --privileged and just using the capabilities for both the bind-mount- and the cvmfsexec-based commands, and said that for cvmfsexec you need to add `--device=/dev/fuse`. I rebuilt the container to get cvmfsexec 4.8 and it works fine.

Do the changes look good?

DrDaveD (changes requested):
Those things are good now but I thought of a couple of more suggestions.

One is that it would probably be good to include singularity.opensciencegrid.org also in the frequently used repos in addition to oasis.opensciencegrid.org.

The other is that you should probably close the CVMFSEXEC_CMDFD file descriptor as the cvmfsexec README advises, to prevent user payload jobs from attempting to mount more repositories.  cvmfsexec is designed to allow mounting more repos from within the entrypoint by executing $CVMFSMOUNT, but the idea was that the pilot wrapper would do it, and your pilot wrapper does not so it's probably too risky to leave it open.  I tried to find an easy way for you to do it and the easiest thing I found that would properly handle the quoting was to create a separate executable file wrapper.sh with one line
```
exec {CVMFSEXEC_CMDFD}>&- "$@"
```
and to change the `"$@"` on the last line of entrypoint.sh to be instead `wrapper.sh "$@"`.

matyasselmeci:
What if I edit `condor_master_wrapper` to add
```
if [[ -n $CVMFSEXEC_CMDFD ]]; then
    exec {CVMFSEXEC_CMDFD}>&-
fi
```
right before the `exec /usr/sbin/condor_master -f` line? Will that let me avoid making another wrapper?

DrDaveD:
Unfortunately that doesn't work because CVMFSEXEC_CMDFD is only set by cvmfsexec for the environment of the command it runs.  I experimented with a bunch of ways to try to get it into entrypoint.sh itself but everything messed up the quoting, for example if there's a blank in one of the parameters.  If you really don't want the extra file I could add an option to cvmfsexec to disable that feature.

Pilot scripts have a similar problem with wanting to execute some code inside singularity before the payload job.  The solutions I have seen used is to have singularity re-execute the same script that invokes singularity but behave differently based on an environment variable.  That's another option to creating a separate file, but more complicated.

matyasselmeci:
Yeah if it's not too hard please add the option to cvmfsexec.

DrDaveD:
Ok there's now a -N option in cvmfsexec-4.10.

matyasselmeci:
The -N option now makes cvmfsexec lock up after
```
CernVM-FS: loading Fuse module... done
CernVM-FS: mounted cvmfs on /cvmfsexec/dist/cvmfs/config-osg.opensciencegrid.org
```
Control-C'ing it gives me
```
^Ccvmfsexec smoke test failed.  You may not have the permissions to run cvmfsexec; see https://github.com/cvmfs/cvmfsexec#README for details
```

DrDaveD:
It's not clear to me how that could happen, and I can't reproduce it interactively.  What's the full command line you test with so I can hopefully reproduce it?

Anyway there isn't a need to use -N on the smoke test, but maybe you'll experience the same problem on the later invocation, I don't know.

matyasselmeci:
Yeah I added it to the smoke test because it failed on the later invocation as well.

Here's my command:
```
podman run --rm -it \
         --cap-add=DAC_OVERRIDE \
         --cap-add=DAC_READ_SEARCH \
         --cap-add=SETUID \
         --cap-add=SETGID \
         --cap-add=SYS_ADMIN \
         --cap-add=SYS_CHROOT \
         --cap-add=SYS_PTRACE \
         --device=/dev/fuse \
         -u 1000:1000 \
         -e CVMFSEXEC_REPOS=oasis.opensciencegrid.org \
          \
         matyasosg/osgvo-docker-pilot:cvmfsexec \
         bash
```
I just pushed a new image.

Looking at entrypoint.sh, what's being run is
```
tar -xzf /cvmfsexec.tar.gz -C /cvmfsexec
/cvmfsexec/cvmfsexec -N -- /bin/true
```

DrDaveD:
I ran that with docker on el7 and got different errors, resolved only by adding option `--security-opt seccomp=unconfined`.  I had noticed that earlier and mentioned it on slack but it should also get into the instructions here when adding cvmfsexec.  That is implied by `--privileged` and I thought it was implied by `--cap-add=SYS_ADMIN` as well but apparently it is not, at least on docker.  It's also needed for singularity so I think  it should be part of your singularity instructions as well.

However, once I added `--security-opt seccomp=unconfined` everything worked and I could not reproduce your problem.  It sounds like it is something specific to Fedora or podman, and I don't have a Fedora VM.  I have podman on my el7 VM, but it doesn't work properly and I get stuck on other problems.  Can you try it with docker on el7?  That's the primary platform we need to support.

DrDaveD:
I was able to reproduce the hang under podman on rocky8.

DrDaveD:
Alright this problem wasn't only in podman or docker, it was podman or docker without `--security-opts systempaths=unmasked`, which I was using when I tested with docker.  Fixed in cvmfsexec 4.11.

matyasselmeci:
Yep, can confirm that it's fixed.

matyasselmeci:
Anything else?

DrDaveD:
I think the cvmfsexec integration is good, although I think with docker it requires `--security-opt seccomp=unconfined`.   setuid-root singularity should be able to get away without it, but not cvmfsexec.  I cannot easily confirm at the moment because I don't think you have updated your matyasosg/osgvo-docker-pilot:cvmfsexec container on dockerhub with the latest cvmfsexec version.

Also I think singularity -cip will not work with the docker instructions you have, and that might hit you when the pilot tries to run a job under singularity.  In my testing I had to add `--security-opt systempaths=unconfined` with all those `--cap-adds` in order to unmask /proc, which singularity -p requires, I believe even when running setuid root.

matyasselmeci:
OK, I just updated the image.

Let's keep this PR focused on cvmfsexec and put any changes for singularity in a separate PR.

matyasselmeci:
I verified that I get errors for `docker` without `--security-opt=seccomp=unconfined` -- I see
```
pivot_root: failed to change root from `.' to `.old-root': Operation not permitted
mkdir: cannot create directory '/cvmfs': Permission denied
mount: mount point /cvmfs/config-osg.opensciencegrid.org does not exist
umount: /cvmfs/config-osg.opensciencegrid.org: mountpoint not found
rmdir: failed to remove '/cvmfs/config-osg.opensciencegrid.org': No such file or directory
```
so I updated the readme to mention that. Strange that it's not needed for podman.

DrDaveD:
Using your current matyasosg/osgvo-docker-pilot:cvmfsexec container if I try to run singularity I get an error 
```
ERROR  : No setuid installation found, for unprivileged installation use: ./mconfig --without-suid
```
Does it work for you?  This is under docker on centos7.  Using singularity -d indicates it's a problem with creating a socketpair:
```
DEBUG   [U=1000,P=593]     init()                        Create socketpair for master communication channel
DEBUG   [U=1000,P=593]     init()                        Create RPC socketpair for communication between stage 2 and RPC server
ERROR   [U=1000,P=593]     init()                        No setuid installation found, for unprivileged installation use: ./mconfig --without-suid
```

We sure are going through a lot of trouble to avoid --privileged.

matyasselmeci:
> Using your current matyasosg/osgvo-docker-pilot:cvmfsexec container if I try to run singularity I get an error
> 
> ```
> ERROR  : No setuid installation found, for unprivileged installation use: ./mconfig --without-suid
> ```
> 
> Does it work for you? This is under docker on centos7. Using singularity -d indicates it's a problem with creating a socketpair:
> 
> ```
> DEBUG   [U=1000,P=593]     init()                        Create socketpair for master communication channel
> DEBUG   [U=1000,P=593]     init()                        Create RPC socketpair for communication between stage 2 and RPC server
> ERROR   [U=1000,P=593]     init()                        No setuid installation found, for unprivileged installation use: ./mconfig --without-suid
> ```

Is this a new issue or does it also happen with, say `opensciencegrid/osgvo-docker-pilot:release`? If it's not new then I'd like to solve it in a separate PR.

> We sure are going through a lot of trouble to avoid --privileged.

I defer to @brianhlin and @rynge on why it was done that way, or if admins will accept just using --privileged and calling it a day.

README.md
```diff
@@ -17,12 +17,22 @@ In order to successfully start payload jobs:
...
+Example invocation utilizing a token for authentication and bind-mounting CVMFS:
 
 ```
 docker run -it --rm --user osg \
        --cap-add=DAC_OVERRIDE --cap-add=SETUID --cap-add=SETGID \
-       --cap-add=CAP_DAC_READ_SEARCH \
+       --cap-add=DAC_READ_SEARCH \
        --cap-add=SYS_ADMIN --cap-add=SYS_CHROOT --cap-add=SYS_PTRACE \
        -v /cvmfs:/cvmfs:shared \
        -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org
```
DrDaveD, line 38:
```suggestion
       -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org \
```

README.md
```diff
@@ -77,4 +87,72 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+```
+docker run -it --rm --user osg \
+       --cap-add=DAC_OVERRIDE --cap-add=SETUID --cap-add=SETGID \
+       --cap-add=DAC_READ_SEARCH \
+       --cap-add=SYS_ADMIN --cap-add=SYS_CHROOT --cap-add=SYS_PTRACE \
+       --device=/dev/fuse \
+       --security-opt=seccomp=unconfined \
+       -v /var/cache/cvmfsexec:/cvmfsexec-cache \
+       -v /var/log/cvmfsexec:/cvmfsexec-logs \
+       -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org
```
DrDaveD, line 150:
```suggestion
       -v /path/to/token:/etc/condor/tokens-orig.d/flock.opensciencegrid.org \
```

DrDaveD:
Even with your container it doesn't fail if I pass in /cvmfs from the host.  It only fails following the cvmfsexec instructions.  I'm not sure why.  singularity exec -u for unprivileged works, except of course not with -p.  If I exchange all the --cap-adds and --security-opts for --privileged, singularity also works privileged.

matyasselmeci:
That's bizarre. OK, I'll use --privileged for the cvmfsexec invocation.

matyasselmeci:
@DrDaveD I changed the instructions for cvmfsexec to use `--privileged`. Does it look good now?

README.md
```diff
@@ -17,15 +17,25 @@ In order to successfully start payload jobs:
...
-Example invocation utilizing a token for authentication:
+In addition, you will be able to run more OSG jobs if you provide CVMFS.  You can do this
+in two ways:
+
+1. Mount CVMFS on the host and bind-mount it into the container by adding `-v /cvmfs:/cvmfs:shared`.
+   This is the preferred mechanism.
+2. Use cvmfsexec as described in [the cvmfsexec section below](#cvmfs-without-a-bind-mount-using-cvmfsexec).
+
+Note: Supporting Singularity jobs inside the container will require the capabilities
+DAC_OVERRIDE, DAC_READ_SEARCH, SETGID, SETUID, SYS_ADMIN, SYS_CHROOT, and SYS_PTRACE.
```
brianhlin, line 28:
```suggestion
`DAC_OVERRIDE`, `DAC_READ_SEARCH`, `SETGID`, `SETUID`, `SYS_ADMIN`, `SYS_CHROOT`, and `SYS_PTRACE`.
```

README.md
```diff
@@ -77,4 +87,68 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+## CVMFS Without a Bind-Mount Using cvmfsexec
+
+If you don't have CVMFS available on the host, the container can still make
+CVMFS available by using [cvmfsexec](https://github.com/cvmfs/cvmfsexec#readme).
+
+You will need to specify a list of CVMFS repos to mount in the environment
+variable `CVMFSEXEC_REPOS`.
+
+On EL7, you must have kernel version >= 3.10.0-1127 (run `rpm -q kernel` to check),
+and user namespaces enabled.  On EL8, you must have kernel version >= 4.18.
```
brianhlin, line 99: For user namespaces, maybe we should point people to step 1 here https://opensciencegrid.org/docs/worker-node/install-singularity/#enabling-unprivileged-singularity?

README.md
```diff
@@ -77,4 +87,68 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+See the cvmfsexec README linked above for details.
```
brianhlin, line 100:
```suggestion
See the [cvmfsexec README](https://github.com/cvmfs/cvmfsexec#readme) details.
```

May as well save people from having to hunt for the link

README.md
```diff
@@ -77,4 +87,68 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+`/cvmfsexec/default.local`; note that options in environment variables are preferred
+over options in `/cvmfsexec/default.local`.
+
+You can store the cache outside of the container by bind-mounting a directory
+to `/cvmfs-cache`.
+You can store the logs outside of the container by bind-mounting a directory to
+`/cvmfs-logs`.
+
+cvmfsexec requires the additional options `--device=/dev/fuse` and
+(for Docker) `--security-opt=seccomp=unconfined`.  Supporting Singularity
```
brianhlin, line 134: I don't understand what this parenthetical is referring to

README.md
```diff
@@ -77,4 +87,68 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+jobs when using cvmfsexec requires privileged containers (`--privileged`).
+
+Here is an example invocation using a token for authentication, using cvmfsexec
+to mount the OASIS and singularity repos instead of bind-mounting `/cvmfs`,
+sending the cache to `/var/cache/cvmfsexec` and the logs to
+`/var/log/cvmfsexec`, and adding support for Singularity jobs:
```
brianhlin, line 140: We should clarify that these paths are on the host

brianhlin (approved):
Some minor comments but otherwise LGTM!

README.md
```diff
@@ -77,4 +87,74 @@ $ singularity build osgvo-pilot.sif docker://opensciencegrid/osgvo-docker-pilot
...
+over options in `/cvmfsexec/default.local`.
+
+You can store the cache outside of the container by bind-mounting a directory
+to `/cvmfs-cache`.
+You can store the logs outside of the container by bind-mounting a directory to
+`/cvmfs-logs`.
+
+cvmfsexec requires the additional options `--device=/dev/fuse` and
+`--security-opt=seccomp=unconfined`.  Supporting Singularity jobs when using
+cvmfsexec requires privileged containers (`--privileged`).
```
DrDaveD, line 139:
```suggestion
cvmfsexec requires privileged containers (`--privileged`).  Using privileged containers avoids the need for those two additional options and all the `--cap-add` options without adding much risk so it is recommended.
```

DrDaveD (approved):
Approved after my one suggested addition to the documentation.

PR #10, opened 2021-03-29
Start expression fixups

brianhlin:
Noticed these when troubleshooting Chameleon cloud's idle slots. Perhaps related: Lancium is also seeing their backfill containers sitting idle.

@rynge what sets `MY.OSG_NODE_VALIDATED`? It doesn't appear to be set in the Chameleon container or one of the JetStream containers:

```
$ podman run --rm -it opensciencegrid/hosted-ce:release condor_status -pool flock.opensciencegrid.org -const '!isUndefined(IsOsgVoContainer) && isUndefined(osg_node_validated)' -af name
slot1@JetStream-1858bc205145
slot1@chameleon-uc-dev-backfill-383906f438aa
```

### DIFF
--- a/50-main.config
+++ b/50-main.config
@@ -51,9 +51,7 @@ GLIDEIN_ResourceName = "Docker"
 STARTD_ATTRS = $(STARTD_ATTRS) GLIDEIN_Country GLIDEIN_Site GLIDEIN_ResourceName WorkerGroupName IsOsgVoContainer
 
 # only accept new work for a while, for validated nodes and jobs with project name set
-# the QDate is do shield from a OSGConnect upgrade and can be removed after 1/1/21
-START = (TARGET.QDate >= 1605631000) && \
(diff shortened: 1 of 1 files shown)


rynge:
@brianhlin It comes from the validation scripts: https://github.com/opensciencegrid/osg-flock/blob/master/node-check/osgvo-node-advertise#L1188 

The idea is to not run any jobs until the node is fully tested.

rynge (approved):
LGTM

brianhlin:
Separately, why do we have a hard limit for accepting jobs? I imagine that in many cases sites will want to accept jobs as long as the containers are running.

rynge:
It is just a recent job limit - used to mitigate black holes (but a pretty poor solution to it).

rynge:
Or do yo mean the time limit? There are two reasons: 1) on a batch system we need to be able to control the lifetime to match the wall time of the jobs, and on a container pool, we want the containers to exit and refresh themselves (could be pulling a new version of the image) automatically.

PR #11, opened 2021-04-14
Mount /cvmfs with :shared; without it, the host can't see automounted volumes

matyasselmeci:
@DrDaveD says not having this might cause issues.


brianhlin (approved):
LGTM!

PR #12, opened 2021-04-21
Use uuidgen for the NETWORK_HOSTNAME in case hostname isn't unique enough

matyasselmeci:


matyasselmeci:
This wouldn't have solved the problem we were running into.

PR #15, opened 2021-05-19
Update the README with instructions for using the /pilot dir

matyasselmeci:


brianhlin:
Should we just remove the README documentation now that we have https://opensciencegrid.org/docs/resource-sharing/os-backfill-containers/ and https://opensciencegrid.org/docs/resource-sharing/user-containers/?

matyasselmeci:
Dunno. I'm adding more stuff to the README in another PR (the cvmfsexec PR). Note that this is the text that shows up on the Docker Hub page for the container.

brianhlin:
> Note that this is the text that shows up on the Docker Hub page for the container.

Huh...how? If that's the case, we should replace the contents of the README with the links to our docs

matyasselmeci:
How about we merge this so it's correct for now, and then we can move the contents to the other doc later (at the next doc focus, perhaps)?

matyasselmeci:
(The README here and the other doc are somewhat different so they will take some time to merge; I already updated the singularity instructions in the other doc but I'd rather not have wrong instructions here, even in the short term.)

brianhlin:
As long as the other docs have the new invocation then that works for me

matyasselmeci:
Yep, updated.

PR #16, opened 2021-06-23
Enable GPUs

rynge:


brianhlin:
I built and pushed a new tag with these changes `rc.testing.pr16.20210623`.

brianhlin:
The Lancium folks have verified that they're advertising GPUs now with the RC image in FD#66546

PR #17, opened 2021-06-25
Delete any whitespace in CVMFSEXEC_REPOS

matyasselmeci:
This lets the user declare the repo list on multiple indented lines like
```
docker run ... \
     -e CVMFSEXEC_REPOS="
            repo1,
            repo2,
            repo3" \
    ...
```

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -9,6 +9,9 @@ cvmfsexec_root=/cvmfsexec
 cvmfsexec_tarball=/cvmfsexec.tar.gz
 cvmfsexec_local_config=$cvmfsexec_root/dist/etc/cvmfs/default.local
 
+# Delete any whitespace in CVMFSEXEC_REPOS
+CVMFSEXEC_REPOS=$(tr -d $' \t\n' <<<"$CMFSEXEC_REPOS")
+
 if [[ -d /cvmfs/config-osg.opensciencegrid.org ]]; then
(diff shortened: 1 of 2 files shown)


matyasselmeci:
I tested the command by hand in bash; `repo` is down so I can't build the container to do a full test.

entrypoint.sh
```diff
@@ -9,6 +9,9 @@ cvmfsexec_root=/cvmfsexec
 cvmfsexec_tarball=/cvmfsexec.tar.gz
 cvmfsexec_local_config=$cvmfsexec_root/dist/etc/cvmfs/default.local
 
+# Delete any whitespace in CVMFSEXEC_REPOS
+CVMFSEXEC_REPOS=$(tr -d $' \t\n' <<<"$CMFSEXEC_REPOS")
```
brianhlin, line 13: What's the `$` do here?

matyasselmeci: `$'...'` interprets backslashed characters:
```
$ echo x\tx
xtx
$ echo x\\tx
x\tx
$ echo 'x\tx'
x\tx
$ echo "x\tx"
x\tx
$ echo $'x\tx'
x	x
```
not to be confused with `$"..."` which translates the message to the current language.

matyasselmeci:
CVMFSEXEC_REPOS is space-separated, not comma-separated. Dunno why I thought the latter.

PR #20, opened 2021-07-01
Added the node validation script. Note that this is fatal to the cont…

rynge:
…ainer if any of the tests fail.


brianhlin:
Ugh, this causes a failure in the Singularity case due to not enough disk space unless we set `--pwd /pilot` and bind mount a volume into `/pilot`

brianhlin (approved):
LGTM now!

PR #25, opened 2021-08-06
Ignore /tokens/ directory which is the default location for tokens obtained from the OS registry

matyasselmeci:


brianhlin (approved):
Sure

PR #26, opened 2021-08-06
Add ways to limit container resources

matyasselmeci:
Add environment variables and hold policy for limiting HTCondor CPUs/memory; also add instructions for limiting via `docker run` arguments.
(SOFTWARE-4634)


matyasselmeci:
Tested this via podman on my laptop; the NUM_CPUS and MEMORY arguments limited the advertised memory (according to log messages). I needed to use `sudo podman` to get the `--memory` option to do anything but that's not surprising.

PR #27, opened 2021-08-11
shut down an idle pilot

dsschult:
Currently `STARTD_NOCLAIM_SHUTDOWN` is fixed at 30 minutes, and supervisord restarts condor repeatedly, never exiting the container.

This PR allows the idle time to be configured, and shuts down the container if `condor_master` exits normally.

### DIFF
--- a/10-htcondor.conf
+++ b/10-htcondor.conf
@@ -1,6 +1,6 @@
 [program:condor_master]
 command=/usr/sbin/condor_master_wrapper
-autorestart=True
+autorestart=unexpected
 startsecs=60
 stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
--- a/10-htcondor.conf
+++ b/10-htcondor.conf
@@ -1,6 +1,6 @@
 [program:condor_master]
 command=/usr/sbin/condor_master_wrapper
-autorestart=unexpected
+autorestart=True
 startsecs=60
 stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -9,7 +9,7 @@ fi
 
 
 #
-# Taken from creation/web_base/condor_startup.sh in the gwms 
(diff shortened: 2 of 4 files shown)


brianhlin:
@dsschult could you rebase this on `master`? We've got some basic CI tests now so I'd like to run this change through them.

dsschult:
Rebased.  And now I see tests running.

brianhlin:
@matyasselmeci @rynge what do you think about this? This would change the behavior of the containers quite a bit for our other users so we'd have to communicate it to them

dsschult:
The main behavior change is `autorestart=unexpected`.  Not sure if that could be made an option, so default behavior is unchanged.  Might be able to do that in the entrypoint script, before supervisord starts up.

brianhlin:
At this point users just start up the container when they want work from the access points and whether or not they get it is our problem. With this change, sites will have to monitor whether or not their container is running to account for transient lulls in usage.

matyasselmeci:
This seems similar to https://opensciencegrid.atlassian.net/browse/SOFTWARE-4608.

brianhlin:
That's in a similar vein but I think it means we only want `autorestart` on successful exit: the container should exit if the startd isn't working so that the admin can just check if their containers are running or not to determine if they're offering their resources to the OSPool.

matyasselmeci:
Wouldn't that delete the log files of that pilot, making it harder to troubleshoot?

dsschult:
I added a change to make it configurable.  Then you can decide which policy you like better at runtime.

brianhlin (approved):
> I added a change to make it configurable. Then you can decide which policy you like better at runtime.

All disagreements can be satisfied by more configuration :). This works for me!

> Wouldn't that delete the log files of that pilot, making it harder to troubleshoot?

We do tell folks to mount something to `/pilot` so the logs should still be around. I think it's probably better than the alternative where you need to be an expert in condor and this container to realize that you're running the container but it's not advertising back to the OSPool.

PR #28, opened 2021-08-20
Moved from the old osg-flock scripts to the current GWMS singularity … (SOFTWARE-4725)

rynge:
…ones

### DIFF
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -7,6 +7,87 @@ if [ `id -u` = 0 ]; then
     exit 1
 fi
 
+
+#
+# Taken from creation/web_base/condor_startup.sh in the gwms 
+# Set a variable read from a file
+#
+pstr='"'
+set_var() {
+    var_name=$1
+    var_type=$2
+    var_def=$3
+    var_condor=$4
+    var_req=$5
+    var_exportcondor=$6
+    var_user=$7
+
+    if [ -z "$var_name" ]; then
+        # empty line
+        return 0
+    fi
+
+    var_val=`grep "^$var_name " $glidein_config | awk '{if (NF>1) ind=length($1)+1; v=substr($0, ind); print substr(v, index(v, $2))}'`
+    if [ -z "$var_val" ]; then
+        if [ "$var_req" == "Y" ]; then
+            # needed var, exit with error
+            #echo "Cannot extract $var_name from '$config_file'" 1>&2
+            STR="Cannot extract $var_name from '$glidein_config'"
+            "$error_gen" -error "condor_startup.sh" "Config" "$STR" "MissingAttribute" "$var_name"
+            exit 1
+        elif [ "$var_def" == "-" ]; then
+            # no default, do not set
+            return 0
+        else
+            eval var_val=$var_def
+        fi
+    fi
+    if [ "$var_condor" == "+" ]; then
+        var_condor=$var_name
+    fi
+    if [ "$var_type" == "S" ]; then
+        var_val_str="${pstr}${var_val}${pstr}"
+    else
+        var_val_str="$var_val"
+    fi
+
+    # insert into condor_config
+    echo "$var_condor=$var_val_str" >> $PILOT_CONFIG_FILE
+
+    if [ "$var_exportcondor" == "Y" ]; then
+        # register var_condor for export
+        if [ -z "$glidein_variables" ]; then
+           glidein_variables="$var_condor"
+        else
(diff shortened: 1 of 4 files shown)


10-setup-htcondor.sh
```diff
@@ -133,3 +217,51 @@ echo
...
+
+# set some defaults for the glideinwms based scripts
+cat >$glidein_config <<EOF
+ADD_CONFIG_LINE_SOURCE $PWD/add_config_line.source
+ALLOW_NONCVMFS_IMAGES True
+CONDOR_VARS_FILE $condor_vars_file
+ERROR_GEN_PATH $PWD/error_gen.sh
+GLIDEIN_SINGULARITY_REQUIRE OPTIONAL
+GLIDEIN_Singularity_Use PREFERRED
+OSG_DEFAULT_CONTAINER_DISTRIBUTION 70%__opensciencegrid/osgvo-el7:latest 30%__opensciencegrid/osgvo-el8:latest
```
dsschult, line 243: I assume this is setting the default singularity container, which will get used if singularity is available and the job does not specify an image.

Is there a way to make this configurable?

rynge: We could expose it as another env var. Or do you have another suggestion on how to interface with container configuration?

dsschult: Yes, env vars work for me.  That would allow me to pick el7 100% of the time.

rynge: Ok, it is now picked up as an env var. You should be able to do:

``
export OSG_DEFAULT_CONTAINER_DISTRIBUTION="100%__opensciencegrid/osgvo-el7:latest"
```

The image part can be anything available under `/cvmfs/singularity.opensciencegrid.org/`

File: 10-setup-htcondor.sh [hunk 2]
```diff
@@ -7,6 +7,87 @@ if [ `id -u` = 0 ]; then
     exit 1
 fi
 
+
+#
+# Taken from creation/web_base/condor_startup.sh in the gwms
```
matyasselmeci, line 12: I guess we have to do this for now but we should try and get upstream to move this function to a sourceable `condor_lib.sh` file, like how they have `singularity_lib.sh`.

10-setup-htcondor.sh
```diff
@@ -133,3 +217,54 @@ echo
...
+export glidein_config=$LOCAL_DIR/glidein_config
+export condor_vars_file=$LOCAL_DIR/main/condor_vars.lst
+
+# set some defaults for the glideinwms based scripts
+if [[ -z $OSG_DEFAULT_CONTAINER_DISTRIBUTION ]]; then
+    OSG_DEFAULT_CONTAINER_DISTRIBUTION="70%__opensciencegrid/osgvo-el7:latest 30%__opensciencegrid/osgvo-el8:latest"
+fi
+cat >$glidein_config <<EOF
+ADD_CONFIG_LINE_SOURCE $PWD/add_config_line.source
+ALLOW_NONCVMFS_IMAGES True
```
matyasselmeci, line 241: This one **must** be configurable, and off by default, at least until we document it. As we're seeing, non-CVMFS images use a lot of disk and network, and the site admin has no way of restricting this.

rynge: Yeah, that is probably a mistake - let me take another test pass

rynge: Fixed. I also added the latest stashcp, as that is an important component if /cvmfs it not available.

10-setup-htcondor.sh
```diff
@@ -133,3 +217,60 @@ echo
...
+OSG_DEFAULT_CONTAINER_DISTRIBUTION $OSG_DEFAULT_CONTAINER_DISTRIBUTION
+SINGULARITY_IMAGE_RESTRICTIONS None
+EOF
+touch $condor_vars_file
+
+# grab the latests copy of stashcp
+mkdir -p client
+
+if (curl --silent --fail --location --connect-timeout 30 --speed-limit 1024 -o client/stashcp http://stash.osgconnect.net/public/dweitzel/stashcp/current/stashcp) &>/dev/null; then
+    chmod 755 client/stashcp
```
matyasselmeci, line 254: What puts "client" in the PATH?

rynge: That is part of the expected gwms directory structure. We just have to mimic it here to make it all work together.

matyasselmeci (approved):
OK, let's do this!

PR #29, opened 2021-08-26
Add tests for validating Singularity within the container (SOFTWARE-4726)

brianhlin:
Test results: https://github.com/brianhlin/osgvo-docker-pilot/actions/runs/1172121470

IIRC, these are the results that we're expecting because cvmfsexec needs unpriv singularity and the current wrapper is affected by this issue: https://opensciencegrid.atlassian.net/browse/SOFTWARE-4725

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -49,7 +49,6 @@ jobs:
       with:
         context: .
         push: false
-        context: .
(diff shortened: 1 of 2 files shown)


brianhlin:
Had to go back to the drawing board on the tests because of impending changes in #28. Will open a new PR

PR #30, opened 2021-08-27
Remove duplicate documentation

brianhlin:


matyasselmeci (approved):
Does what it says on the can.

brianhlin:
Note that these docs contain instructions for how to limit resources that are not in the official docs. We've got a ticket for adding the missing bits here https://opensciencegrid.atlassian.net/browse/SOFTWARE-4789

PR #33, opened 2021-09-14
Ensure that Nvidia libs are bind mounted into child containers

brianhlin:
(SOFTWARE-4807)

Work around for https://github.com/hpcng/singularity/issues/5759, courtesy of @DrDaveD 

We do this to ensure that GPU libs can be accessed by user containers.
In particular, the backfill container mounts the libs from the host to
'/.singularity.d/libs' (available in LD_LIBRARY_PATH).

ldconfig tries to write its cache to /etc/, which isn't possible if
the outer backfill container is started up with Singularity.

### DIFF
--- a/10-ldconfig-cache.sh
+++ b/10-ldconfig-cache.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+# Rebuild the ld.so cache for Singularity child containers (SOFTWARE-4807)
+ldconfig "$LD_LIBRARY_PATH"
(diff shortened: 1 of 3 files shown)


brianhlin:
I tried Singularity-in-Singularity in the CHTC GPULab and this seems to work as far as I can tell. Instead of the image init script, we could just change the wrapper to generate `/pilot/ld.so.cache` if it doesn't already exist

rynge (approved):
Looks great!

PR #34, opened 2021-09-15
Add container tag to startd ad (SOFTWARE-4602)

brianhlin:


Dockerfile
```diff
@@ -81,7 +82,11 @@ COPY 10-cleanup-htcondor.sh /etc/osg/image-cleanup.d/
 COPY 10-htcondor.conf /etc/supervisord.d/
 COPY 50-main.config /etc/condor/config.d/
 RUN chmod 755 /bin/entrypoint.sh
- 
+
+RUN [[ -n $TIMESTAMP ]] || TIMESTAMP=$(date +%Y%m%d-%H%M); \
```
matyasselmeci, line 86: Not sure how `docker build` handles `ARG` expansion; did you try out this build with an empty `TIMESTAMP`?

brianhlin: Yeah, this works but I'm revisiting this since we should separate manual builds out from our automated builds.

Dockerfile
```diff
@@ -83,8 +83,12 @@ COPY 10-htcondor.conf /etc/supervisord.d/
...
 RUN chmod 755 /bin/entrypoint.sh
 
-RUN [[ -n $TIMESTAMP ]] || TIMESTAMP=$(date +%Y%m%d-%H%M); \
-    sed -i "s|@CONTAINER_TAG@|opensciencegrid/osgvo-docker-pilot:${BASE_YUM_REPO}-${TIMESTAMP}|" \
+RUN if [[ -n $TIMESTAMP ]]; then \
+       tag=${BASE_YUM_REPO}-${TIMESTAMP} \
+     else \
+       tag=manual-build \
+    fi; \
+    sed -i "s|@CONTAINER_TAG@|opensciencegrid/osgvo-docker-pilot:$tag|" \
```
matyasselmeci, line 91: I don't like unconditionally putting `opensciencegrid/osgvo-docker-pilot` in the tag. What if I make my own for testing and it's `matyasosg/osgvo-docker-pilot`?

brianhlin: Good point. I'll just set the attr to the empty string if the TIMESTAMP build arg isn't specified

50-main.config
```diff
@@ -49,7 +49,9 @@ GLIDEIN_Country = "US"
 GLIDEIN_Site = "Docker"
 GLIDEIN_ResourceName = "Docker"
 
-STARTD_ATTRS = $(STARTD_ATTRS) GLIDEIN_Country GLIDEIN_Site GLIDEIN_ResourceName WorkerGroupName IsOsgVoContainer
+GLIDEIN_ContainerTag = @CONTAINER_TAG@
```
matyasselmeci, line 52:
```suggestion
GLIDEIN_ContainerTag = "@CONTAINER_TAG@"
```
Dunno if quotes are necessary here but the other attributes have them.

matyasselmeci (approved):

PR #35, opened 2021-09-15
Test reduced capabilities (SOFTWARE-4477)

brianhlin:
Unsurprisingly, this works and now we can test it!

EDIT: Reading over the ticket, it sounds like we can try reduced caps for the `cvmfsexec` case now that we moved over to the GlideinWMS Singularity scripts in #28

### DIFF
--- a/tests/test_inside_gha.sh
+++ b/tests/test_inside_gha.sh
@@ -212,7 +212,8 @@ case "$CVMFS_INSTALL" in
         start_cvmfs
         ;;
     cvmfsexec)
-        DOCKER_EXTRA_ARGS=(-e CVMFSEXEC_REPOS='oasis.opensciencegrid.org singularity.opensciencegrid.org')
+        DOCKER_EXTRA_ARGS=(-e CVMFSEXEC_REPOS='oasis.opensciencegrid.org singularity.opensciencegrid.org'
+                           --device=/dev/fuse)
         ;;
 esac
 
--- a/tests/test_inside_gha.sh
+++ b/tests/test_inside_gha.sh
@@ -204,13 +204,9 @@ CONTAINER_RUNTIME="$1"
 CVMFS_INSTALL="$2"
 case "$CVMFS_INSTALL" in
     bindmount)
-        DOCKER_EXTRA_ARGS=(--cap-add DAC_OVERRIDE
-                           --cap-add DAC_READ_SEARCH
-                           --cap-add SETUID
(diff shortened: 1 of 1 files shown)


matyasselmeci:
Dunno about "unsurprisingly" but hey, not complaining. So did you check that the the two CVMFS volumes got mounted (i.e. exist and are non-empty)?

brianhlin:
@matyasselmeci looks like the `cvmfsexec` test cases failed. The reduced caps on the bind-mounted CVMFS look, good though: by checking the value of `HAS_SINGULARITY`, we're also testing CVMFS availability within the container.

brianhlin:
@DrDaveD we added some CI tests to verify some bits of the backfill container and we're trying to test the following capabilities: https://github.com/opensciencegrid/osgvo-docker-pilot/pull/35/files#diff-3ce2aeee6e5389638907be79b09253b67574fb2151dab8ce4d2f68d2852d5cf1R8-R10. These work fine for the bindmount case but don't for `cmvfsexec`. Do you have thoughts on what capabilities would work now that we've moved the containers over to the GlideinWMS Singularity scripts?

DrDaveD:
I think you're just missing `--device /dev/fuse` which is also recommended in the [install-singularity](https://opensciencegrid.org/docs/worker-node/install-singularity/#enabling-unprivileged-singularity) documentation.

brianhlin:
@matyasselmeci @rynge tests look happy now!

tests/test_inside_gha.sh
```diff
@@ -5,6 +5,10 @@ OSP_TOKEN_PATH=/tmp/token
 COMMON_DOCKER_ARGS="run --user osg
                         --detach
                         --security-opt apparmor=unconfined
+                        --security-opt seccomp=unconfined
+                        --security-opt systempaths=unconfined
+                        --security-opt no-new-privileges
+                        --net=host
```
matyasselmeci, line 11: Switching to host networking is a big change; is that something we're doing just for GHA or are we going to be telling people to do it for their pilots?

brianhlin: We started doing it here https://github.com/opensciencegrid/docs/pull/874

matyasselmeci: Okay, then.
Might want to explain in that document why you are doing it but that's a separate PR.

matyasselmeci (approved):

PR #36, opened 2021-09-21
Ensure that we run per-job module setup (SOFTWARE-4819, OSPOOl-6)

brianhlin:
Also, move GlideinWMS script dir structure creation to the Dockerfile. Still testing this change on Tiger.

### DIFF
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -214,6 +214,7 @@ export GLIDEIN_Site="$GLIDEIN_Site"
 export GLIDEIN_ResourceName="$GLIDEIN_ResourceName"
 export OSG_SITE_NAME="$GLIDEIN_ResourceName"
 export OSG_SQUID_LOCATION="$OSG_SQUID_LOCATION"
+export GWMS_DIR="$LOCAL_DIR"
 exec $LOCAL_DIR/condor_job_wrapper.sh "\$@"
 EOF
 chmod 755 $LOCAL_DIR/user-job-wrapper.sh
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -236,7 +236,7 @@ echo
 cd $LOCAL_DIR
 
 # gwms files in the correct location
-cp -a /gwms/. .
+cp -a /gwms/. $LOCAL_DIR/
 cp -a /usr/sbin/osgvo-singularity-wrapper condor_job_wrapper.sh
 
 # minimum env to get glideinwms scripts to work
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -235,11 +235,7 @@ echo
 cd $LOCAL_DIR
 
 # gwms files in the correct location
-cp -a /gwms/* .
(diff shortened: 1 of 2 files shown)


brianhlin:
This did not fix the problem. Back to the drawing board

brianhlin:
After some debugging, the culprit appeared to be the executable bit on the `setup_prejob.sh` script, which other GlideinWMS scripts will take as an indication that the script should be executed instead of sourced.

brianhlin:
Tested and verified using Tiger:

```
[blin@login05 ~]$ cat module-test.sh
#!/bin/bash

hostname

module -v 2>&1

env | grep MODULE
[blin@login05 ~]$ cat module-test.out
osgvo-docker-pilot-74cbf6954d-t5pss

Modules based on Lua: Version 7.8  2018-07-23 18:20 -05:00
    by Robert McLay mclay@tacc.utexas.edu

HAS_MODULES=True && HAS_CVMFS_oasis_opensciencegrid_org && HAS_CVMFS_connect_opensciencegrid_org
MODULEPATH=/cvmfs/connect.opensciencegrid.org/modules/modulefiles/linux-rhel7-x86_64/Core
MATLAB_COMPATIBLE=HAS_MODULES && HAS_FILE_usr_lib64_libXt_so_6
MODULESHOME=/cvmfs/connect.opensciencegrid.org/modules/packages/linux-rhel7-x86_64/gcc-4.8.5spack/lmod-7.8-gd44e35n2z5a6zbe54bpjtc55wrvjity/lmod/lmod
```

10-setup-htcondor.sh
```diff
@@ -235,11 +236,7 @@ echo
 cd $LOCAL_DIR
 
 # gwms files in the correct location
-cp -a /gwms/* .
-mkdir -p .gwms.d/bin
-for target in cleanup  postjob  prejob  setup  setup_singularity; do
-    mkdir -p .gwms.d/exec/$target
-done
+cp -a /gwms/. $LOCAL_DIR/
```
edquist, line 239: I guess that's an interesting way to avoid putting things under `$LOCAL_DIR/gwms/`, right?

Looks like it'll do the trick.

brianhlin: Well that and `cp -a /gwms/*` was not picking up the `.gwms*` dirs

edquist: Ah, is that (picking up the `.gwms*` dirs) what was meant by "Ensure that we run per-job module setup" ?

brianhlin: That's part of it, the other bit that enables this is `/gwms/.gwms.d/exec/prejob/setup_prejob.sh`, which was added in the same commit

edquist: Ah right, with the curl command.  Very trusting of those glideinWMS folks ;)

Thanks for the explanation.

edquist (approved):
I have not carefully analyzed the `setup_prejob.sh` from `glideinWMS` that we are pulling in (and trusting), but Brian says it worked for him and the changes in this repo look to work.

PR #37, opened 2021-09-28
Update Dockerfile

bbockelm:
(Whoops, forgot a usable commit message)

The `attr` RPM is needed to test for CVMFS functionality.


brianhlin:
We opted for #38 instead

PR #38, opened 2021-09-28
Install attr for CVMFS error detection (SOFTWARE-4834)

brianhlin:
Same as #37 with a commit message attached to a Jira ticket


brianhlin:
@bbockelm approved this change offline. Merging.

PR #39, opened 2021-10-03
Small Dockerfile tweaks

matyasselmeci:
- Install git along with the other RPMs to avoid creating a yum cache
- Download osgvo scripts in a separate layer than the RPM installs


brianhlin (approved):
LGTM

PR #44, opened 2021-11-30
Moved to OSG 3.6, and updated config to point to the new OSPool central managers

rynge:


matyasselmeci (approved):
LGTM. @brianhlin , what do you think?

brianhlin:
Yup, I'm on board. I'd like to eventually move to EL8 as well but that should probably be a more controlled and tested migration.

PR #45, opened 2021-12-06
Avoid accidentally printing $TOKEN on the console

matyasselmeci:
- turn off xtrace (set -x) when using $TOKEN in an expression
- don't use it on the command line, pipe it through cat instead
- set it to a bogus value once we're done with it


rynge (approved):
Looks good

PR #48, opened 2022-01-07
Log *all* condor logs to stdout

bbockelm:
This (a) adds a syslog daemon to the container (via a setuid binary
as syslog is immovably at `/dev/log`) and (b) tells all HTCSS daemons
to log to the container.

The syslog daemon is set to log to the supervisord stdout, resulting
in all HTCondor logs getting to the container's stdout in a timely
manner.

### DIFF
--- a/generate-hostcert
+++ b/generate-hostcert
@@ -39,6 +39,7 @@ with open(sys.argv[1], "rb") as fp:
 
 response = requests.post("https://os-registry.osgdev.chtc.io/syslog-ca/issue", data={"csr": csr_bytes},
     headers={"Authorization": f"Bearer {token.decode()}"},
+    timeout=20,
 )
 
 response_dict = response.json()
--- a/generate-hostcert
+++ b/generate-hostcert
@@ -1,0 +1,53 @@
+#!/usr/bin/python3
+
+import os
+import sys
+import json
+
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography.hazmat.backends import default_backend
+
+from cryptography import x509
+from cryptography.x509.oid import NameOID
+from cryptography.hazmat.primitives import hashes
+
+import requests
+
+# First, generate a new host key.
+key = rsa.generate_private_key(
+    public_exponent=65537,
+    key_size=2048,
+    backend=default_backend(),
+)
+
+key_bytes = key.private_bytes(
+    encoding=serialization.Encoding.PEM,
+    format=serialization.PrivateFormat.TraditionalOpenSSL,
+    encryption_algorithm=serialization.NoEncryption()
+)
+
+csr = x509.CertificateSigningRequestBuilder(
+   ).subject_name(x509.Name([x509.NameAttribute(NameOID.COMMON_NAME, u"will_be_overwritten")])
+   ).sign(key, hashes.SHA256(), backend=default_backend())
+
+
+csr_bytes = csr.public_bytes(serialization.Encoding.PEM).decode()
+
+with open(sys.argv[1], "rb") as fp:
+    token = fp.read().strip()
+
+response = requests.post("https://os-registry.osgdev.chtc.io/syslog-ca/issue", data={"csr": csr_bytes},
+    headers={"Authorization": f"Bearer {token.decode()}"},
+)
+
+response_dict = response.json()
+response_dict["key"] = key_bytes.decode()
+
+os.makedirs("/etc/pki/rsyslog", exist_ok=True)
+with open("/etc/pki/rsyslog/ca.crt", "w") as fp:
+    fp.write(response_dict["ca"])
+with open("/etc/pki/rsyslog/tls.crt", "w") as fp:
+    fp.write(response_dict["certificate"])
+with open("/etc/pki/rsyslog/tls.key", "w") as fp:
+    fp.write(response_dict["key"])
--- a/generate-hostcert
+++ b/generate-hostcert
@@ -38,7 +38,7 @@ with open(sys.argv[1], "rb") as fp:
     token = fp.read().strip()
 
 try:
-    response = requests.post("https://os-registry.osgdev.chtc.io/syslog-ca/issue", data={"csr": csr_bytes},
+    response = requests.post("https://os-registry.opensciencegrid.org/syslog-ca/issue", data={"csr": csr_bytes},
         headers={"Authorization": f"Bearer {token.decode()}"},
         timeout=20,
     )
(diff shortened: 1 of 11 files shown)


dsschult:
Is there a reason why `launch_rsyslogd` is compiled from src instead of a shell script?

matyasselmeci:
`launch_rsyslogd` is setuid root so it can run `rsyslogd` as root, and modern Linuxes do not allow setuid shell scripts.

Leaving GCC in the image does bloat it by ~50 MB so it might be better if `launch_rsyslogd` were pre-compiled and then copied in. Something like

```
FROM alpine:latest AS compile
COPY launch_rsyslogd.c /tmp/launch_rsyslogd.c
RUN apk --no-cache add gcc musl-dev && \
 cc -static -o /launch_rsyslogd /tmp/launch_rsyslogd.c && \
 strip /launch_rsyslogd

FROM opensciencegrid/software-base:3.6-el7-${BASE_YUM_REPO}
...
COPY --from=compile /launch_rsyslogd /usr/bin/launch_rsyslogd
RUN chmod 04755 /usr/bin/launch_rsyslogd
...
```

bbockelm:
I _knew_ if I put `gcc` in the main image it'd nerdsnipe someone into doing the multistage build for me.  Thanks @matyasselmeci!

bbockelm:
Ok -- if you now check `/var/log` in the `central-syslog` pod on `osgdev`, you'll see the forwarded logs from the pods as well...

dsschult:
> `launch_rsyslogd` is setuid root so it can run `rsyslogd` as root, and modern Linuxes do not allow setuid shell scripts.

Thanks, I didn't know that.  Makes sense from a security perspective.

@bbockelm Can you make the remote logging configurable?  Not sure you want logging from IceCube glideins.

bbockelm:
> Can you make the remote logging configurable? Not sure you want logging from IceCube glideins.

This shouldn't be a problem -- you need to get one of the generated certificates to send to the syslog server.  None of the IceCube glideins should have one!

bbockelm:
Ok - in the latest update, we gather the output from supervisord, all startup activities, and HTCondor and get it to the same stream of data.

I now parse the non-standard HTCondor timestamp format and use that in the rsyslog messages.  This includes structured data as well (for example, the slot name):

```
<149>1 2022-01-09T21:11:09.460Z f5ca0a4dcfab condor_startd 2025 - [cat="D_ALWAYS"] ERROR: SECMAN:2010:Received "DENIED" from server for user brianb@flock.opensciencegrid.org using method IDTOKENS.
<149>1 2022-01-09T21:11:11.896Z f5ca0a4dcfab condor_starter 2027 - [cat="D_ALWAYS" slot="testing"] my_popenv: Failed to exec /usr/bin/java, errno=2 (No such file or directory)
<149>1 2022-01-09T21:11:11.896Z f5ca0a4dcfab condor_starter 2027 - [cat="D_ALWAYS" slot="testing"] JavaDetect: failed to execute /usr/bin/java -classpath /usr/share/condor:. CondorJavaInfo old
<149>1 2022-01-09T21:11:12.673Z f5ca0a4dcfab condor_starter 2027 - [cat="D_ALWAYS" slot="testing"] Failed to read results from '/usr/bin/docker -v': 'No such file or directory' (2)
<149>1 2022-01-09T21:11:12.673Z f5ca0a4dcfab condor_starter 2027 - [cat="D_ALWAYS" slot="testing"] DockerAPI::detect() failed to detect the Docker version; assuming absent.
<150>1 2022-01-09T21:11:09.222Z f5ca0a4dcfab condor_master 2017 - [cat="D_ALWAYS"] Daemons::StartAllDaemons all daemons were started
```

Note I also now dynamically generate part of the syslog configuration; remote syslog is now done only if a specific environment variable (`$SYSLOG_HOST`) is set.

brianhlin (approved):
Minor suggestions but otherwise looks reasonable. I'm definitely interested in the implementation on the dev registry side but I assume that previously issued certs will work even if we move the syslog/registry to prod.

supervisord_startup.sh
```diff
@@ -0,0 +1,27 @@
...
+exec 3>&1 4>&2 &> >(tee -a /tmp/startup.log)
+
+# Allow the derived images to run any additional runtime customizations
+shopt -s nullglob
+for x in /etc/osg/image-init.d/*.sh; do source "$x"; done
+shopt -u nullglob
+
+# Allow child images to add cleanup customizations
+function source_cleanup {
+    for x in /etc/osg/image-cleanup.d/*.sh; do source "$x"; done
```
matyasselmeci, line 14: probably should do the `nullglob` thing here too

generate-hostcert
```diff
@@ -0,0 +1,54 @@
...
+csr_bytes = csr.public_bytes(serialization.Encoding.PEM).decode()
+
+with open(sys.argv[1], "rb") as fp:
+    token = fp.read().strip()
+
+response = requests.post("https://os-registry.osgdev.chtc.io/syslog-ca/issue", data={"csr": csr_bytes},
+    headers={"Authorization": f"Bearer {token.decode()}"},
+    timeout=20,
+)
+
```
matyasselmeci, line 44:
```suggestion
try:
    response = requests.post("https://os-registry.osgdev.chtc.io/syslog-ca/issue", data={"csr": csr_bytes},
        headers={"Authorization": f"Bearer {token.decode()}"},
        timeout=20,
    )
except OSError as err:
    print("*** Error downloading certificate from registry: %s" % err, file=sys.stderr)
    raise
```

generate-hostcert
```diff
@@ -0,0 +1,54 @@
...
+response_dict = response.json()
```
matyasselmeci, line 45:
```suggestion
try:
    response_dict = response.json()
except ValueError as err:
    print("*** Bad data received from registry: %s" % err, file=sys.stderr)
    print("*** Received data: %s", response, file=sys.stderr)
    raise
```

matyasselmeci (approved):

generate-hostcert
```diff
@@ -0,0 +1,62 @@
...
+   ).sign(key, hashes.SHA256(), backend=default_backend())
+
+
+csr_bytes = csr.public_bytes(serialization.Encoding.PEM).decode()
+
+with open(sys.argv[1], "rb") as fp:
+    token = fp.read().strip()
+
+try:
+    response = requests.post("https://os-registry.osgdev.chtc.io/syslog-ca/issue", data={"csr": csr_bytes},
```
brianhlin, line 41:
```suggestion
    response = requests.post("https://os-registry.opensciencegrid.org/syslog-ca/issue", data={"csr": csr_bytes},
```

Once we deploy the new registry to prod, we should point to it

matyasselmeci: I don't think we have the opensciencegrid.org alias so we should change this to `osg.chtc.io` for now

brianhlin: Hrm,we're getting it from somewhere! https://os-registry.opensciencegrid.org/

matyasselmeci: Never mind, I was thinking of the syslog pod.

matyasselmeci: New registry is up.

PR #49, opened 2022-01-08
Add filebeats to the container for log collection

djw8605:
Completely untested, but it builds 

Alternative to #48


brianhlin:
For the time being, we've decided to continue with #48 as to not add the filebeats RPM dep (it'd be nice if it were in EPEL!) and to avoid hooking this up into the critical GRACC infrastructure

PR #53, opened 2022-01-21
Don't set up remote syslog if we can't find SYSLOG_HOST

matyasselmeci:
- Fix indent
- Don't set up remote syslog if we can't find SYSLOG_HOST


bbockelm (approved):
LGTM.

PR #54, opened 2022-01-21
Clean out the rsyslog directory prior to starting.

bbockelm:
Avoids issues if the /pilot directory is persistent

### DIFF
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -153,6 +153,10 @@ set -x
 export _CONDOR_SEC_PASSWORD_FILE=$LOCAL_DIR/condor/tokens.d/flock.opensciencegrid.org
 export _CONDOR_SEC_PASSWORD_DIRECTORY=$LOCAL_DIR/condor/passwords.d
 
+# There's at least one site that makes `/pilot` a persistent volume; we want to remove
+# the cruft from the prior run.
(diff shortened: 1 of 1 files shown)


10-setup-htcondor.sh
```diff
@@ -153,6 +153,10 @@ set -x
 export _CONDOR_SEC_PASSWORD_FILE=$LOCAL_DIR/condor/tokens.d/flock.opensciencegrid.org
 export _CONDOR_SEC_PASSWORD_DIRECTORY=$LOCAL_DIR/condor/passwords.d
 
+# There's at least one site that makes `/pilot` a persistent volume; we want to remove
+# the cruft from the prior run.
+rm -rf /pilot/{log,rsyslog}
```
matyasselmeci, line 158: These should be under $LOCAL_DIR instead of /pilot in case there are multiple pilots sharing the same volume.

bbockelm: Yes -- but that's a much larger surgery.  Would like to get Purdue working for now and then fixup the rest of the approach later.

matyasselmeci (approved):

10-setup-htcondor.sh
matyasselmeci: Fair enough.

PR #55, opened 2022-01-25
Enable syslog streaming by default.

bbockelm:


matyasselmeci (approved):
Turn on the firehose!
Do we have log rotation on the central syslog servers?

bbockelm:
Yup, we logrotate daily and have this on a separate 100GB partition... we'll see whether that suffices.  From the current volumes from the test sites, should be good.

PR #56, opened 2022-01-30
Make regexp more permissive.

bbockelm:
On failures (and in a few other situations), HTCondor will log categories along the lines of `(D_ALWAYS|D_FAILURE)`; add the pipe character (`|`) to the regexp to account for this situation.


brianhlin (commented):
Not sure what type of regexes syslog uses: do `|` have to be escaped?

matyasselmeci:
Not inside a character class.

brianhlin (approved):
Good point, LGTM then

PR #57, opened 2022-02-02
Use the single-binary stashcp as stash_plugin

matyasselmeci:
This is a copy, not a symlink, because the program uses argv[0] to determine its behavior.


brianhlin (approved):
LGTM!

PR #59, opened 2022-02-02
Sanitize GLIDEIN_ResourceName before using it as NETWORK_HOSTNAME

matyasselmeci:
Get rid of illegal characters, chop off all leading/trailing dots/dashes, truncate so that NETWORK_HOSTNAME <= 63 chars


rynge (approved):
Looks good!

PR #61, opened 2022-02-03
ITB backfill container (SOFTWARE-5002)

matyasselmeci:
https://opensciencegrid.atlassian.net/browse/SOFTWARE-5002

If you specify `--build-arg ITB=1` when building the container, it will use the itb versions of the osg-flock scripts and also add the attributes for joining the ITB pool.

Other features (intended for manual builds):
- specify your fork of osg-flock in the `OSG_FLOCK_REPO` var, e.g. `--build-arg OSG_FLOCK_REPO=matyasselmeci/osg-flock`
- specify the branch of osg-flock in the `OSG_FLOCK_BRANCH` var, e.g. `--build-arg OSG_FLOCK_BRANCH=mybranch`
- the above two args, as well as `OSG_FLOCK_HASH` (the git hash of the checkout) will be advertised in the startd ad
- specify `--build-arg RANDOM=$RANDOM` to use the docker build cache for the RPM install, but get a fresh checkout of the scripts. This saves a lot of time, especially on slow connections.

The -development-itb tag gets built in the workflow; I don't build -release-itb and -testing-itb because the build matrix is already pretty big.

Includes #60.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -118,7 +118,7 @@ COPY rsyslog.conf /etc/
 RUN chmod 755 /bin/entrypoint.sh
 
 RUN if [[ -n $TIMESTAMP ]]; then \
-       tag=opensciencegrid/osgvo-docker-pilot:${BASE_YUM_REPO}-${TIMESTAMP}; \
+       tag=opensciencegrid/osgvo-docker-pilot:${BASE_YUM_REPO}${ITB+-itb}-${TIMESTAMP}; \
     else \
        tag=; \
     fi; \
--- a/Dockerfile
+++ b/Dockerfile
@@ -119,6 +119,13 @@ RUN if [[ -n $TIMESTAMP ]]; then \
     sed -i "s|@CONTAINER_TAG@|$tag|" \
            /etc/condor/config.d/50-main.config
 
+RUN \
+    if [[ -n $ITB ]]; then \
+        echo 'Is_ITB_Site = True'  >> /etc/condor/config.d/55-itb.config; \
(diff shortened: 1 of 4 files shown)


rynge (commented):
Should it maybe also register to the ITB pool? I don't have strong feelings about this as you can also control access via the START expression.

matyasselmeci:
Sure, what do I need to do for that?

matyasselmeci:
Is that what pilots do? Does that mean you'll have to do `condor_status -pool cm-1.ospool-itb.osg-htc.org` to see them?

matyasselmeci:
OK, I changed it to use the ITB CMs by default, but you can turn it off by setting `ITBPOOL=0` in the environment.

Dockerfile
```diff
@@ -125,6 +125,8 @@ RUN if [[ -n $TIMESTAMP ]]; then \
 
 RUN \
     if [[ -n $ITB ]]; then \
+        # Set the default pool to ITB, but allow turning off with -e ITBPOOL=0
+        echo 'export ITBPOOL=${ITBPOOL:-1}' > /etc/osg/image-init.d/01-itb.sh; \
```
brianhlin, line 129: Any reason you're doing this instead of just setting `ENV ITBPOOL=-1` in the Dockerfile?

rynge (approved):
Looks good

brianhlin (approved):
Could you change the variable to `ITB_OSPOOL`? We may have an ITB PATh facility or something in the future! One small comment but otherwise LGTM

Dockerfile
matyasselmeci: I only want to set the default of ITBPOOL if the `ITB` build arg is set, and Docker doesn't really have `if` statements.

PR #62, opened 2022-02-03
Install sshd for condor_ssh_to_job; the service doesn't need to be started up, we just need the software

matyasselmeci:
(SOFTWARE-4970)


rynge (approved):
lgtm

timtheisen:
Why is this needed? The condor RPM requires openssh-server already.

matyasselmeci:
I just checked and Tim's right.  So we don't need this.

PR #63, opened 2022-02-07
Add prod PATh facility choice to POOL (SOFTWARE-5004)

brianhlin:
Verified that condor started up (I didn't have OSPool tokens handy) when setting `POOL` to nothing, `prod-ospool`, and `itb-ospool` were set. @ddavila0 is working on cm-2 for the PATh facility so that doesn't seem to work quite yet

### DIFF
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -277,7 +277,9 @@ elif [[ $POOL =~ (itb|prod)-ospool ]]; then
     CCB_SUFFIX="9619?sock=collector$(random_range 1 6)"
 fi
 
-# Append the CCB suffix to each host in CONDOR_HOST
+# Append the CCB suffix to each host in CONDOR_HOST, e.g.
+# "cm.school.edu:10576", or
+# "cm-1.ospool.osg-htc.org:9619?sock=collector6,cm-2.ospool.osg-htc.org:9619?sock=collector6"
 if [[ -n $CCB_RANGE_LOW && -n $CCB_RANGE_HIGH ]] ||
        [[ $POOL =~ (itb|prod)-ospool ]]; then
     CCB_ADDRESS=$(python -Sc "import re; \
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -168,15 +168,6 @@ rm -rf /pilot/{log,rsyslog}
 mkdir -p /pilot/{log,log/log,rsyslog,rsyslog/pid,rsyslog/workdir,rsyslog/conf}
 touch /pilot/log/{Master,Start,Proc,SharedPort,XferStats,log/Starter}Log /pilot/log/StarterLog{,.testing}
 
-# Configure remote peer if applicable
-if [[ "x$SYSLOG_HOST" == "x" ]]; then
-    if is_true "$ITBPOOL"; then
-        SYSLOG_HOST="syslog.osgdev.chtc.io"
-    else
-        SYSLOG_HOST="syslog.osg.chtc.io"
-    fi
-fi
-
 # Set some reasonable defaults for the token registry.
 if [[ "x$REGISTRY_HOST" != "x" ]]; then
     REGISTRY_HOSTNAME="$REGISTRY_HOST"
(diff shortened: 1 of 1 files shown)


10-setup-htcondor.sh
```diff
@@ -228,23 +219,64 @@ EOF
...
+
+# Don't assume a CCB unless they specify CCB_RANGE_LOW or CCB_RANGE_HIGH
+if [[ -n $CCB_RANGE_LOW ]] || [[ -n $CCB_RANGE_HIGH ]]; then
     CCB_PORT=$(python -S -c "import random; print(random.randrange($CCB_RANGE_LOW,$CCB_RANGE_HIGH+1))")
-    CCB_ADDRESS="\$(CONDOR_HOST):$CCB_PORT"
-else
-    CCB_COLLECTOR=$(python -S -c "import random; print(random.randrange(1,6))")
-    CCB_ADDRESS="${OSPOOL_CM1}:9619?sock=collector$CCB_COLLECTOR,${OSPOOL_CM2}:9619?sock=collector$CCB_COLLECTOR"
+    if [[ $POOL =~ (itb|prod)-ospool ]]; then
+        CCB_PORT="9619?sock=collector$CCB_PORT"
```
matyasselmeci, line 275: Whatever this is, it is not a port.

I'm really having trouble following this code or figuring out what comes out of it. IMO it's OK to repeat yourself if it enhances clarity.

10-setup-htcondor.sh
```diff
@@ -228,23 +219,64 @@ EOF
...
+# POOL CONFIGURATION #
+######################
+
+# Default to the production OSPool
+POOL=${POOL:=prod-ospool}
+
+case ${POOL} in
+    itb-ospool)
+        CCB_RANGE_LOW=1
+        CCB_RANGE_HIGH=5
```
matyasselmeci, line 233: Where did `CCB_RANGE_LOW` and `CCB_RANGE_HIGH` come from before this change?

10-setup-htcondor.sh
```diff
@@ -228,23 +219,64 @@ EOF
...
 
-# extra HTCondor config
-# if CCB_RANGE_* is set, use the old config, otherwise assume OSPool with shared port
-if [[ "x$CCB_RANGE_LOW" != "x" ]]; then
+# Configure remote peer if applicable
+SYSLOG_HOST=${SYSLOG_HOST:-$default_syslog_host}
+
+
+# Don't assume a CCB unless they specify CCB_RANGE_LOW or CCB_RANGE_HIGH
+if [[ -n $CCB_RANGE_LOW ]] || [[ -n $CCB_RANGE_HIGH ]]; then
```
matyasselmeci, line 272:
```suggestion
# Don't assume a CCB unless they specify CCB_RANGE_LOW and CCB_RANGE_HIGH
if [[ -n $CCB_RANGE_LOW && -n $CCB_RANGE_HIGH ]]; then
```

10-setup-htcondor.sh
brianhlin: We didn't set it before per se but the values came from https://github.com/opensciencegrid/osgvo-docker-pilot/pull/63/files#diff-8dbb461a6753f09dffe3e11c79e48c6e7b194d2392514fd4024b7fbbbd435b5aL245

10-setup-htcondor.sh
brianhlin: Does the most recent commit make it clearer?

matyasselmeci: Not really, and it was more readable when you set the CCB low/high in the `case` statement.

Perhaps you could add an example of what the string is expected to look like?

brianhlin: Added and just reworked the tortured logic. Should be clearer now.

10-setup-htcondor.sh
```diff
@@ -228,23 +225,66 @@ EOF
...
-if [[ "x$CCB_RANGE_LOW" != "x" ]]; then
-    CCB_PORT=$(python -S -c "import random; print(random.randrange($CCB_RANGE_LOW,$CCB_RANGE_HIGH+1))")
-    CCB_ADDRESS="\$(CONDOR_HOST):$CCB_PORT"
-else
-    CCB_COLLECTOR=$(python -S -c "import random; print(random.randrange(1,6))")
-    CCB_ADDRESS="${OSPOOL_CM1}:9619?sock=collector$CCB_COLLECTOR,${OSPOOL_CM2}:9619?sock=collector$CCB_COLLECTOR"
+# Configure remote peer if applicable
+SYSLOG_HOST=${SYSLOG_HOST:-$default_syslog_host}
+
+if [[ -n $CCB_RANGE_LOW && $CCB_RANGE_HIGH ]]; then
```
matyasselmeci, line 270:
```suggestion
if [[ -n $CCB_RANGE_LOW && -n $CCB_RANGE_HIGH ]]; then
```

10-setup-htcondor.sh
```diff
@@ -228,23 +225,66 @@ EOF
...
+    # e.g., cm.school.edu:10576
+    CCB_SUFFIX=$(random_range "$CCB_RANGE_LOW" "$CCB_RANGE_HIGH")
+elif [[ $POOL =~ (itb|prod)-ospool ]]; then
+    # Choose a random OSPool collector for CCB
+    # e.g., cm-1.ospool.osg-htc.org:9619?sock=collector3
+    CCB_SUFFIX="9619?sock=collector$(random_range 1 6)"
 fi
+
+# Append the CCB suffix to each host in CONDOR_HOST
+if [[ -n $CCB_RANGE_LOW && $CCB_RANGE_HIGH ]] ||
```
matyasselmeci, line 281:
```suggestion
if [[ -n $CCB_RANGE_LOW && -n $CCB_RANGE_HIGH ]] ||
```

10-setup-htcondor.sh
```diff
@@ -228,23 +225,66 @@ EOF
...
+       [[ $POOL =~ (itb|prod)-ospool ]]; then
+    CCB_ADDRESS=$(python -Sc "import re; \
+print(','.join([cm + ':$CCB_SUFFIX' \
+for cm in re.split(r'[\s,]+', '$CONDOR_HOST')]))")
```
matyasselmeci, line 285: Can you add an example of what this looks like?

brianhlin: Done!

matyasselmeci: Thanks!

matyasselmeci (approved):

PR #64, opened 2022-02-11
Ensure that the container shuts down correctly after idle/retirement

rynge:
I noticed that the current container did not shut down after exhausting the `ACCEPT_IDLE_MINUTES`/`ACCEPT_JOBS_FOR_HOURS` timeouts. HTCondor would exit, but restart by supervisord. This PR introduces a `DEFAULT_MASTER_SHUTDOWN_SCRIPT` script which runs `kill 1`, similar to the annex HTCondor documentation: https://htcondor.readthedocs.io/en/latest/cloud-computing/annex-customization-guide.html

### DIFF
--- a/10-htcondor.conf
+++ b/10-htcondor.conf
@@ -1,6 +1,6 @@
 [program:condor_master]
 command=/usr/sbin/condor_master_wrapper
-autorestart=True
+autorestart=False
 startsecs=60
 stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
--- a/10-htcondor.conf
+++ b/10-htcondor.conf
@@ -1,6 +1,6 @@
 [program:condor_master]
 command=/usr/sbin/condor_master_wrapper
-autorestart=False
+autorestart=unexpected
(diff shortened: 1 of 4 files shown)


dsschult:
This is why I run with the env variable `SUPERVISORD_RESTART_POLICY="unexpected"`.  If HTCondor stops normally, it will exit.  But maybe you want more control?

rynge:
Yes, I think this is important as the expectations when using `ACCEPT_IDLE_MINUTES`/`ACCEPT_JOBS_FOR_HOURS` is not correct. We could discuss the defaults of those values, but when specified the container has to shut down correctly.

matyasselmeci:
You might want to do `autorestart=unexpected` instead of `autorestart=False` so it still tries running condor again in case of error. That shouldn't interfere with your ACCEPT_IDLE_MINUTES/ACCEPT_JOBS_FOR_HOURS because the shutdown script kills supervisord itself.

rynge:
Ok, I changed autostart to `unexpected `

PR #65, opened 2022-02-25
Relax GPU start/rank expressions so that a container will run jobs ev…

rynge:
…en if there are no GPU jobs in the pool, but still prefer GPU jobs if available

### DIFF
--- a/50-main.config
+++ b/50-main.config
@@ -53,14 +53,15 @@ GLIDEIN_ContainerTag = "@CONTAINER_TAG@"
 STARTD_ATTRS = $(STARTD_ATTRS) GLIDEIN_Country GLIDEIN_Site GLIDEIN_ResourceName WorkerGroupName IsOsgVoContainer GLIDEIN_ContainerTag
 
 # only accept new work for a while, for validated nodes and jobs with project name set
-# On GPU nodes, CPU only jobs are only allowed after all GPUs are in use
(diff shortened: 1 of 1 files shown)


matyasselmeci (approved):
LGTM.

dsschult:
I personally dislike this policy, since it's very likely for CPU jobs to steal all the GPUs in a pool, and then they won't be available when a GPU job comes along.  At least, we tried this initially with IceCube's private pool, and quickly moved to reserve a core for each GPU and only run GPU jobs on those.

matyasselmeci:
Can we add preemption of CPU jobs by GPU jobs perhaps?

rynge:
Preemption is certainly a solution - I think it might be disabled pool wide of the OSPool, but I'm happy to work on that bit. 

Another solution would be to create a flag the admin can set to control which behavior they prefer.

Yet another option would be to have "flavors" of the image, that is different tags for the different behaviors. I think this would be confusing. I would lean towards the flag shortterm, and the preemption longterm.

dsschult:
> I would lean towards the flag shortterm, and the preemption longterm.

I'd agree to that.

rynge:
Ok, I will make a new PR

bbockelm:
One trick we have used before is to have the slot go to CPU jobs only if the GPU has remained idle for 15 minutes.

PR #66, opened 2022-02-26
Added flag ALLOW_CPUJOB_ON_GPUSLOT to give admins control over if the…

rynge:
…y want to make gpu slots available to cpu-only jobs

Note that the default for the flag is to keep the same functionality as we have now.

@dsschult Let me know if this is good by you.


dsschult (approved):
LGTM

matyasselmeci (approved):

PR #68, opened 2022-03-22
Also set CCB for the prod PATh facility

brianhlin:


matyasselmeci (approved):

matyasselmeci:
Just mash retry until it passes!

Actually all the bind-mount tests (and only the bind-mount tests) are failing.

brianhlin:
Good catch on it being limited to bind mounts. Turns out the CVMFS install is failing so I've updated the tests so they bail earlier so it's easier to understand the failures in the future

I've poked Dave about the install issues

brianhlin:
This failure is due to https://sft.its.cern.ch/jira/browse/CVM-2075, so I'm going to merge since this only affects the tests

PR #74, opened 2022-04-05
Aligned START expression with OSPool glideins. Added support for DESI…

rynge:
…RED_Sites/UNDESIRED_Sites (SOFTWARE-4874), IsBlackHole (replaces old RecentJobStarts approach), and HasExcessiveLoad.

### DIFF
--- a/50-main.config
+++ b/50-main.config
@@ -50,6 +50,10 @@ HasExcessiveLoad = LoadAvg > 2*DetectedCpus + 2
 
 # only accept new work for a while, for validated nodes and jobs with project name set
 # On GPU nodes, CPU only jobs are only allowed after all GPUs are in use
+#
+# Some attributes used (OSG_NODE_VALIDATED, IsBlackHole, HasExcessiveLoad, ...) might
+# not be used/defined by all VOs, so we explicitly check if they are defined before
+# using them as part of the START expression.
 START = (isUndefined(DaemonStartTime) || (time() - DaemonStartTime) < (MY.ACCEPT_JOBS_FOR_HOURS*60*60)) && \
         (isUndefined(MY.OSG_NODE_VALIDATED) || MY.OSG_NODE_VALIDATED) && \
         (isUndefined(IsBlackHole) || IsBlackHole == False) && \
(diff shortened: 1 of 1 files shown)


dsschult:
Is there a reason for why the DESIRED_Sites and UNDESIRED_Sites are being matched on the startd instead of in the job requirements?  I'm curious if there are good reasons for this, from either policy or HTCondor performance perspectives (I thought it was more performant to put these in the job requirements).

rynge:
For the glideins, DESIRED_Sites/UNDESIRED_Sites are also used for signaling what sites to provision the glidines for. It is basically a way to target just a subset of sites within a group. The problem for the OSPool is that for DESIRED_Sites/UNDESIRED_Sites to work correctly, all startds have to use them. That is why we need them here on the containers as well.

50-main.config
```diff
@@ -41,14 +41,22 @@ GLIDEIN_ResourceName = "Docker"
...
+               IsBlackHole HasExcessiveLoad
+
+IsBlackHole = IfThenElse(RecentJobDurationAvg is undefined, false, RecentJobDurationCount >= 10 && RecentJobDurationAvg < 180)
+HasExcessiveLoad = LoadAvg > 2*DetectedCpus + 2
 
 # only accept new work for a while, for validated nodes and jobs with project name set
 # On GPU nodes, CPU only jobs are only allowed after all GPUs are in use
 START = (isUndefined(DaemonStartTime) || (time() - DaemonStartTime) < (MY.ACCEPT_JOBS_FOR_HOURS*60*60)) && \
-        (MY.OSG_NODE_VALIDATED == True) && \
+        (isUndefined(MY.OSG_NODE_VALIDATED) || MY.OSG_NODE_VALIDATED) && \
```
matyasselmeci, line 54: This also runs jobs if OSG_NODE_VALIDATED is undefined - is that on purpose? When would OSG_NODE_VALIDATED be undefined?

rynge: The main concern would be other VOs picking up the container and maybe using different validation scripts.

matyasselmeci: Can you add a comment saying that, so I don't ask this question next time?

rynge: Done

matyasselmeci (approved):
LGTM.

PR #75, opened 2022-04-12
Moved image to EL8

rynge:


matyasselmeci:
Looks like you'll have to replace "python" commands with "python3" commands.

matyasselmeci:
Alternatively you could stick
```
alternatives --set python /usr/bin/python3
```
to make unversioned python call python3 instead.

rynge:
Committed a fix for the Python issue.

Another "problem" might be that cvmfsexec is based on OSG 3.5. Do you know if there are any plans on moving it to 3.6?

matyasselmeci:
This should do it: https://github.com/cvmfs/cvmfsexec/pull/42

matyasselmeci (approved):

PR #76, opened 2022-04-20
Use standard OSG container actions 

brianhlin:
- Adds release series specific tags (SOFTWARE-4944)
- Removes ITB tags, these were interfering with flux wildcards anyway

Working build/test here: https://opensciencegrid.atlassian.net/browse/SOFTWARE-4944


brianhlin:
No one else is around! Merging!

allbymyself

PR #78, opened 2022-04-26
OSPOOL-19: Upgrading to HTCondor 9.8.1

rynge:


Dockerfile
```diff
@@ -14,18 +14,19 @@ ARG BASE_OSG_SERIES=3.6
...
-        condor \
         osg-wn-client \
         redhat-lsb-core \
         singularity \
         attr \
         git \
         rsyslog rsyslog-gnutls python3-cryptography python3-requests \
         bind-utils \
+ && yum -y install --enablerepo=osg-upcoming-testing \
+        condor \
```
brianhlin, line 29: Can we wrap this in an `if [[ $BASE_YUM_REPO == "release" ]]`?

brianhlin: Oh and a `# FIXME` comment to remove it after today's release. We may bring it back depending on what we decide here https://opensciencegrid.atlassian.net/browse/OSPOOL-17?focusedCommentId=368979

timtheisen:
I think it safer to install condor from osg-upcoming and the upgrade it from osg-upcoming-testing. Alternatively, we could just wait until the afternoon when condor 9.8.1 has been released and the containers rebuilt.

brianhlin:
> I think it safer to install condor from osg-upcoming and the upgrade it from osg-upcoming-testing.

I'm not sure I see the benefit of this vs installing directly from osg-upcoming-testing. Could you elaborate?

Dockerfile
brianhlin: > Oh and a # FIXME comment to remove it after today's release. We may bring it back depending on what we decide here https://opensciencegrid.atlassian.net/browse/OSPOOL-17?focusedCommentId=368979

On second thought, don't add the FIXME comment. Let's just go with this solution

rynge:
No longer required.

PR #79, opened 2022-04-26
add "want_mpi =!= true" to START expression (SOFTWARE-5048)

edquist:
SOFTWARE-5048 originally called for adding just `want_mpi =!= true` to the `START` expression.

In the comments, @matyasselmeci then suggested `(TARGET.Want_MPI =!= True || MY.Has_MPI =?= True)` might be better, so I've added that instead.


brianhlin:
@edquist can you apply this change to one of the containers in tiger in the `osg` namespace, restart the condor_master service, and submit a job from login04/05 to verify that this expression works as expected?

PR #80, opened 2022-05-02
Disable OSG 3.5 updates (SOFTWARE-5143)

brianhlin:


edquist (approved):
nothing too crazy here

matyasselmeci (approved):
does what it says on the tin

PR #82, opened 2022-05-04
Prefer local singularity over the cvmfs unpriv one - this will increa…

rynge:
…se our ability to use sif images

### DIFF
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -453,7 +453,7 @@ else
 fi
 rm -f /tmp/stashcp-debug.txt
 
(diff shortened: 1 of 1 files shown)


dsschult:
I'm guessing this is explicitly for the case of running the pilot in docker with extra privs?  Vs if the pilot was started in singularity, there should be no advantage.

rynge:
No, the unpriv Singularity is not at good a running sif files (needs to unpack them), so we have that disabled (https://github.com/opensciencegrid/osg-flock/blob/master/node-check/osgvo-default-image#L105-L109). This change just prefers the priv install.

matyasselmeci (approved):
LGTM. Let's give it a shot.

PR #83, opened 2022-05-13
Always install stashcp from the osg-flock repo even if it's already present; overwrite existing copy (SOFTWARE-5192)

matyasselmeci:


brianhlin (approved):
LGTM

PR #84, opened 2022-05-26
Add GLIDEIN_Rank_Extra

matyasselmeci:
Allow adding an expression onto the startd's RANK expression by setting the `GLIDEIN_Rank_Extra` environment variable.


matyasselmeci:
RANK essentially does nothing (see comment in the ticket). Closing.

PR #85, opened 2022-05-27
Don't bother setting RANK

matyasselmeci:
The schedd presents jobs to the negotiator one at a time, and so RANK does nothing without preemption.


rynge:
Let's leave this one sitting until we have had a discussion about enabling preemption on the ospool.

PR #90, opened 2022-06-01
Unset SINGULARITY_BIND so that singularity-in-singularity doesn't attempt to propagate bind mounts from outside (SOFTWARE-5207)

matyasselmeci:
(which leads to 'mount source ... does not exist' errors, see https://github.com/apptainer/singularity/issues/5766)


brianhlin (approved):
LGTM!

PR #91, opened 2022-06-02
Have container put jobs on hold if they allocated disk for the slot.

bbockelm:
Copy/pasted the config knobs into my local test box, seemed to work well.

Only thing of note is that disk usage is scanned once every 5 minutes -- this will only help for 'slow growth' of usage, not for people who unpack tarballs and immediately push nodes out-of-space.


matyasselmeci:
Nope, it's the shell doing it -- see my suggestion.

PR #92, opened 2022-06-06
kill supervisord with SIGQUIT (SOFTWARE-4608)

edquist:
kill defaults to SIGTERM, but I found some pages [1,2] with examples killing supervisord specifically using SIGQUIT.  Could this be the key to what went wrong?

Then again, the supervisord docs suggest[1] that a number of signals including TERM will cause it to shut down, so I am not wildly confident this will actually fix it.

[1] https://serverfault.com/questions/760726/how-to-exit-all-supervisor-processes-if-one-exited-with-0-result
[2] https://hangarau.space/using-supervisord-as-the-init-process-of-a-docker-container/
[3] http://supervisord.org/running.html#signal-handlers

### DIFF
--- a/master_shutdown.sh
+++ b/master_shutdown.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 
 # kill the container when the master shuts down
-kill 1
+kill -QUIT 1
 
--- a/master_shutdown.sh
+++ b/master_shutdown.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 
 # kill the container when the master shuts down
-kill -QUIT 1
+supervisorctl shutdown
 


brianhlin:
I don't think that this will solve the `cvmfsexec` case since I believe that `cvmfsexec` and `supervisord` are two distinct process trees. We test the cvmfsexec and bindmount cases, maybe add some tests to verify that when condor shuts down that it also takes down the container?

edquist:
>I don't think that this will solve the cvmfsexec case since I believe that cvmfsexec and supervisord are two distinct process trees.  We test the cvmfsexec and bindmount cases, maybe add some tests to verify that when condor shuts down that it also takes down the container?

In the cvmfsexec process tree in SOFTWARE-4608's description, i do not see any condor processes.  Is the problem that the condor processes go away in the cvmfsexec case without taking anything else down?

I think i see what you mean, that supervisord is run as a child (descendant) of cvmfsexec, and in that case the `kill 1` won't apply to it.  It _will_ apply to cvmfsexec itself, but only if cvmfsexec has an explicit signal handler for shutting down, as apparently linux protects pid 1 from default signals that would kill the process.

We could kill the pid in `/var/run/supervisord.pid` instead, if that's the pidfile supervisord uses - i think that is the case for root, but i'm not sure about non-root, and in the cvmfsexec case it's run non-root.

But i read somewhere you can run `supervisorctl shutdown`, which sounds a bit less desperate, if it works.

So that's what i'm trying.

matyasselmeci:
We'll merge https://github.com/opensciencegrid/topology/pull/2491 and then you can get yourself a token.

brianhlin (approved):
Huzzah, let's do it!

PR #95, opened 2022-07-21
Use the new CCB servers for the OSPool

rynge:
Also fixed a stashcp issue when stashcp is installed in the image, and a escaping issue when writing out the HTCondor config file.


matyasselmeci (approved):
LGTM.

PR #98, opened 2022-09-23
SOFTWARE-5317 Update OSG container actions to automatically generate the TIMESTAMP_TAG build variable

AidenCohen31:
Updates to action versions may be necessary (EX: lowercasing for load-action is not right version right now)

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -9,15 +9,6 @@ on:
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


brianhlin:
@AidenCohen31 we'll need to tag new versions for each of the actions to make use of them so I'd like to get all the action PRs right first before modifying this GHA workflow

matyasselmeci:
`master` has diverged a bit since this PR was created, especially in #118. Do we still want this?

PR #102, opened 2022-11-01
use carl's fixes for singularity_lib.sh (SOFTWARE-5340)

edquist:
at least until we can get https://github.com/glideinWMS/glideinwms/pull/233 merged upstream


matyasselmeci (approved):

matyasselmeci:
This is fine though instead of mixing and matching gwms scripts from different repos, it would be better to do something like what I did with the osgvo scripts with `OSG_FLOCK_REPO` and `OSG_FLOCK_BRANCH` a few lines further down...

matyasselmeci (approved):
This will result in `add_config_line.source` to be executable but I don't think that's a big deal.

edquist:
>This will result in `add_config_line.source` to be executable

Wait how do you figure?

edquist:
I think it's 644 (ie, not executable) in the git repo, and the chmod 755 only applies to the `*.sh` files, right?

matyasselmeci:
`install(1)` makes it executable unless you pass `-m`:
>        -m, --mode=MODE
>              set permission mode (as in chmod), instead of rwxr-xr-x

edquist:
Oh wow, TIL

PR #109, opened 2022-11-24
Install ITB scripts and allow selecting between production and ITB

matyasselmeci:
This installs both the ITB and production scripts into separate directories. When the container is started with `-e ITB=1`, the ITB scripts will be used and the pilot will advertise itself as an ITB site and join the ITB pool.

This does not yet use `additional-htcondor-config` or the `prepare-job` hook, and it uses the old singularity wrapper for ITB as well, so right now the two kinds of pilots are very similar.  Those fixups will be coming in a later PR.

### DIFF
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -162,6 +162,7 @@ case ${POOL} in
         default_ccb2=ccb-2.ospool-itb.osg-htc.org
         default_syslog_host=syslog.osgdev.chtc.io
         GLIDECLIENT_Group=itb-container
+        itb_sites_start_clause=' && (TARGET.ITB_Sites =?= True)'
         ;;
     prod-ospool)
         default_cm1=cm-1.ospool.osg-htc.org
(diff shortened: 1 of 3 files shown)


dsschult:
Just an FYI that I hadn't noticed it before, but both this and 44afcd70d9f3b3826b1a98048e9c92bab2f7d752 require the pool to be in the list you set, so break IceCube using this container for our own glideins.

matyasselmeci:
https://github.com/opensciencegrid/osgvo-docker-pilot/commit/44afcd70d9f3b3826b1a98048e9c92bab2f7d752 has been in there since September; how have you been using this image so far? Maybe we can add a `POOL=custom` or something...

dsschult:
> [44afcd7](https://github.com/opensciencegrid/osgvo-docker-pilot/commit/44afcd70d9f3b3826b1a98048e9c92bab2f7d752) has been in there since September; how have you been using this image so far? Maybe we can add a `POOL=custom` or something...

Mostly I haven't updated in a while from an older tagged version.  Too busy with other things to work on this :disappointed: 

But yes, a `POOL=custom` would be nice.

matyasselmeci:
OK, that will be a separate PR, but I made a ticket for adding it: https://opensciencegrid.atlassian.net/browse/SOFTWARE-5383

bbockelm (approved):
LGTM.

Only questions is whether we should tinker with the START expression to avoid non-ITB jobs?

bbockelm (approved):
Ok, I like this even better now that we protect against production jobs from running on ITB resources.

PR #110, opened 2022-11-30
Install cvmfsexec sooner in the build process

matyasselmeci:
makedist downloads RPMs and it's good to put RPM installation earlier in the Dockerfile for better caching.


rynge (approved):
LGTM

PR #103, opened 2022-11-16
Use upstream glideinwms repo and master branch again

matyasselmeci:
Our desired changes (glideinWMS#233) got merged

PR #113, opened 2022-12-09
Fix set_var() to not use the variable name as the value if the variable has an empty value

matyasselmeci:
If a variable in the glidein config looked like `OSG_SITE_NAME ` (i.e. with a trailing space but no value), then the awk expression in set_var() would use `OSG_SITE_NAME ` as the value instead of the empty string.

This was fixed in GlideinWMS somewhere between 3.9.6rc4 and 3.9.6 so we should use their version once we can (so after #103 is fixed and merged).

### DIFF
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -27,7 +27,8 @@ set_var() {
         return 0
     fi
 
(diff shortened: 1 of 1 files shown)


rynge (approved):
LGTM

PR #114, opened 2023-01-06
Install "apptainer" instead of "singularity"

matyasselmeci:
The RPM named "singularity" has been removed from EPEL and the "provides: singularity" line has been removed from apptainer, so now we need to ask for it by name.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN useradd osg \
  && yum -y install \
         osg-wn-client \
         redhat-lsb-core \
-        singularity \
+        apptainer \
         attr \
         git \
         rsyslog rsyslog-gnutls python3-cryptography python3-requests \


brianhlin (approved):
LGTM

matyasselmeci:
Hm. Development tests failed with
```
+++ condor_config_val EXECUTE
Configuration Error Line 8 while reading config source /etc/condor/config.d/10-stash-plugin.conf
```
I have a new version of the osdf-client RPMs in development, maybe I messed up the config.

brianhlin:
Interesting, ok. We can still merge this, though, since we pretend that the development containers don't exist https://osg-htc.org/technology/policy/container-release/#tags

matyasselmeci:
I'm about to build a new RPM with a fix

matyasselmeci:
I built an osdf-client RPM with a fix but it's only in minefield, not in development; I'll merge this and re-run the build in a few hours.

PR #115, opened 2023-01-06
Simplify logic for having a custom CONTAINER_TAG

matyasselmeci:


edquist (approved):
Nice!  This looks a lot better.

The replaced code looks way over 80-wide so i assume i didn't have anything to do with it anyway ;)

PR #118, opened 2023-01-10
Fix advertised container tag for new build action

brianhlin:
Otherwise we get junk like this:

`opensciencegrid/osgvo-docker-pilot:3.6-release-opensciencegrid/osgvo-docker-pilot:3.6-cuda_11_8_0-release-20230109-2007`

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -32,7 +32,7 @@ jobs:
           fi
 
       - id: build-image
-        uses: opensciencegrid/build-container-action@v0.5.0
+        uses: opensciencegrid/build-container-action@v0.6.0
         with:
           osg_series: ${{ matrix.osg_series }}
           osg_repo: ${{ matrix.repo }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,9 +22,7 @@ ENV ITB=
 # Previous args have gone out of scope
 ARG BASE_OSG_SERIES=3.6
 ARG BASE_YUM_REPO=testing
-ARG TIMESTAMP_TAG
-
-ARG CONTAINER_TAG=${TIMESTAMP_TAG}
+ARG TIMESTAMP_IMAGE=osgvo-docker-pilot:${BASE_OSG_SERIES}-${BASE_YUM_REPO}
 
 RUN useradd osg \
  && mkdir -p ~osg/.condor \
@@ -153,7 +151,7 @@ COPY 10-htcondor.conf 10-rsyslogd.conf /etc/supervisord.d/
 COPY 50-main.config /etc/condor/config.d/
 COPY rsyslog.conf /etc/
 RUN chmod 755 /bin/entrypoint.sh
-RUN sed -i "s|@CONTAINER_TAG@|${CONTAINER_TAG}|" /etc/condor/config.d/50-main.config
+RUN sed -i "s|@CONTAINER_TAG@|${TIMESTAMP_IMAGE}|" /etc/condor/config.d/50-main.config
 
 
 RUN chown -R osg: ~osg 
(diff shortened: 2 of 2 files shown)


matyasselmeci:
That seems wrong. `${TIMESTAMP_TAG}` should just be the timestamp, why is it the entire image name?

brianhlin:
Ah sure, I changed the build action to give the timestamp image instead. I'll update the action to change the env var name (this is the only workflow that I know of that uses this env var) then update this PR

matyasselmeci (approved):
Kinda weird that an arg called "TIMESTAMP_IMAGE" doesn't have one but ok...

brianhlin:
Added a date, though I think it may actually be clearer if we made the default completely different from the normal format. For example, if we have a custom, local build deployed on Tiger and it shows up in the pool, that should jump out to us in this attr from the startd ad

matyasselmeci:
I did not know you could do shell substitutions in Dockerfiles. I have mixed feelings about that...

if I were making custom images for running on Tiger, I would put them in the `matyasosg` or `osgpreview` projects instead of `opensciencegrid` but it looks like your new expression omits the project name?

edquist:
>I did not know you could do shell substitutions in Dockerfiles

https://docs.docker.com/engine/reference/builder/#run

Looks like if you do `RUN <command>` it runs `/bin/sh -c "<command>"` ... if you use the list version then probably shell variable substitutions won't work.

brianhlin:
> Looks like if you do RUN <command> it runs /bin/sh -c "<command>" ... if you use the list version then probably shell variable substitutions won't work.

Ah yeah, good call. I guess I could move the shell var sub down to where we do the replacement in the condor config but that changes this from a default value for the `TIMESTAMP_IMAGE` to a default value for the `GLIDEIN_ContainerTag` HTCondor config var. They're currently the same thing but it could lead to some confusion down the line.

> if I were making custom images for running on Tiger, I would put them in the matyasosg or osgpreview projects instead of opensciencegrid but it looks like your new expression omits the project name?

I don't think it's possible to know the project (or the container tag) a priori, which honestly makes this build arg a PITA. You have to align your build processes such that you're passing along the FQIN you're using in during the build process, passing along the FQIN constituent parts, and/or the Dockerfile has to make lots of assumptions about FQIN format.

I think the only sensible default value is actually the empty string just so you don't have to specify the `TIMESTAMP_IMAGE` when building locally and that jumps out at us if we do a `condor_status -af GLIDEIN_ContainerTag`.

edquist:
>>Looks like if you do RUN it runs /bin/sh -c "" ... if you use the list version then probably shell variable substitutions won't work.
>
>Ah yeah, good call. I guess I could move the shell var sub down

Eh, what i meant was it should work fine how you have it (in the non-list syntax)

```
RUN sed -i "s|@CONTAINER_TAG@|${TIMESTAMP_IMAGE}|" /etc/condor/config.d/50-main.config
```

This works because the whole line (after the `RUN`) is passed through the shell and thus gets interpreted by the shell (including shell variable substitutions).

What i think _won't_ work is to specify the args as a list

```
RUN ["sed", "-i", "s|@CONTAINER_TAG@|${TIMESTAMP_IMAGE}|", "/etc/condor/config.d/50-main.config"]
```

because these would be literal strings passed to exec(3) rather than passed through the shell.

matyasselmeci:
> I think the only sensible default value is actually the empty string

"UNKNOWN" might jump out at us more :)

brianhlin:
> "UNKNOWN" might jump out at us more :)

"NAUGHTY DEVELOPER"

brianhlin:
> This works because the whole line (after the RUN) is passed through the shell and thus gets interpreted by the shell (including shell variable substitutions).

Ohhh sure, I agree with Mat: I've got mixed feelings on that so I'm going to go with the static `UNKNOWN` instead.

PR #120, opened 2023-01-20
Fix regexp in rsyslog.conf to match categories with `:2`

matyasselmeci:


brianhlin (commented):
Is there any other specification other than `:2`? Couldn't the diff just add `(:2)?`?

matyasselmeci:
Believe it or not, `:1` also exists.
But yeah maybe my regex is too general -- not expecting a `:319` anytime soon :)

brianhlin (approved):
LGTM

PR #121, opened 2023-01-20
Use Alma 8 as the base image for containers

matyasselmeci:


brianhlin (approved):
Testing went well according to the Jira ticket so LGTM

PR #122, opened 2023-01-20
Add ENABLE_REMOTE_SYSLOG (default on)

matyasselmeci:
Add a knob to enable/disable remote syslog (it defaults to enabled). Without this there's no clean way to disable it, since setting SYSLOG_HOST to blank just makes it use the default syslog host.


brianhlin (approved):
LGTM

PR #123, opened 2023-01-26
Ask for apptainer by name (SOFTWARE-5401)

matyasselmeci:
We are already using apptainer for running singularity containers, but we're calling it through the /usr/bin/singularity symlink. Use the binary directly.


brianhlin (approved):
LGTM

PR #124, opened 2023-01-31
Add GLIDEIN_RANDOMIZE_NAME knob to add a random string to the NETWORK_HOSTNAME

matyasselmeci:
[SOFTWARE-5222](https://opensciencegrid.atlassian.net/browse/SOFTWARE-5222)

Add a 10-char random string to the NETWORK_HOSTNAME so it's unique on places using host networking.

Also turns out I had the rules for domain names wrong -- instead of a 63-char length limit overall, it's a 63-char length limit _per dot-separated-component_ and a 255-char length limit overall. This makes generating names a bit simpler.

[SOFTWARE-5222]: https://opensciencegrid.atlassian.net/browse/SOFTWARE-5222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -337,21 +337,20 @@ print(','.join([cm + ':$CCB_SUFFIX' \
 for cm in re.split(r'[\s,]+', '$CONDOR_HOST')]))")
 fi
 
-# https://whogohost.com/host/knowledgebase/308/Valid-Domain-Name-Characters.html rules
-hostname_length=$(hostname | wc -c)
-maxlen=$(( 63 - $hostname_length ))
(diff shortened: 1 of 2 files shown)


matyasselmeci:
Note that the output we send via rsyslog uses the actual hostname instead of anything we come up with.  (It _also_ sends GLIDEIN_ResourceName and GLIDEIN_Site); we should probably change that to use NETWORK_HOSTNAME to correlate outputs better but I'd rather do it in a separate ticket.

brianhlin (approved):
LGTM!

PR #125, opened 2023-02-07
Allow use of the prepare-job-hook for ITB container pilots

matyasselmeci:
Specifying `ITB=1` in the environment will cause the pilot to use the scripts from the `itb` frontend group, now including `additional-htcondor-config`. Config that's done in `additional-htcondor-config` will no longer be done directly in the pilot.

Setting `CONTAINER_PILOT_USE_JOB_HOOK=1` will additionally use the minimal job wrapper and job hook instead of the singularity job wrapper. The default is off but we can turn it on by default when we declare it production ready.

The singularity job wrapper is still used to run the `advertise-userenv` startd_cron.


matyasselmeci:
I haven't tested GPU support yet.

matyasselmeci:
I tested GPU support on an EP on Tiger; works just fine with HTCondor 10.2.2 though I have to set `SINGULARITY_DISABLE_PID_NAMESPACES=1` in the environment so the Glidein's detection of HAS_SINGULARITY doesn't fail.

rynge (approved):
LGTM

PR #126, opened 2023-02-21
SOFTWARE-5299: implemented OSG_PROJECT_RESTRICTION and OSPool tagging

rynge:
Users can now set -e OSG_PROJECT_NAME="..." to restrict what projects the pilot will serve.

Using the old GLIDEIN_Start_Extra will result in a deprecation warning.

OSPool attribute is set accordingly - True if it is an unrestricted pilot to the production OSPool CMs, False otherwise.

### DIFF
--- a/50-main.config
+++ b/50-main.config
@@ -44,7 +44,8 @@ GLIDEIN_ContainerTag = "@CONTAINER_TAG@"
 STARTD_ATTRS = $(STARTD_ATTRS) GLIDEIN_Country GLIDEIN_Site GLIDEIN_ResourceName \
                WorkerGroupName IsOsgVoContainer GLIDEIN_ContainerTag \
                GLIDECLIENT_Group \
-               Is_ITB_Site
+               Is_ITB_Site \
+               START_EXTRA OSG_PROJECT_RESTRICTION
 
 # IsBlackHole is set in "additional-htcondor-config"
 # HasExcessiveLoad is set in "additional-htcondor-config"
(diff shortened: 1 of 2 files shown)


matyasselmeci:
What would it take to do these in the ospool-pilot scripts instead (https://github.com/opensciencegrid/osg-flock/tree/master/ospool-pilot/main/pilot)?  Probably advertise-base since that's where we have existing code.  It would be useful to only have to implement this once.

dsschult:
Note that `GLIDEIN_Start_Extra` is convenient for debugging, when you only want to match a specific job.  While it may not be used by the OSPool, it's a nice convenience feature and doesn't harm anything else.  I'd be sad to see it go away.

matyasselmeci:
A bit hacky but what if we only show the warning if GLIDEIN_Start_Extra contains `ProjectName` (case insensitive)?

rynge:
Good point about advertise-base - I have updated this PR, and submitted https://github.com/opensciencegrid/osg-flock/pull/219

50-main.config
```diff
@@ -65,7 +65,7 @@ START = (isUndefined(DaemonStartTime) || (time() - DaemonStartTime) < (MY.ACCEPT
         ($(CPUJOB_ON_GPUSLOT)) && \
         (isUndefined(TARGET.SingularityImage) || MY.SINGULARITY_START_CLAUSE) && \
         (TARGET.Want_MPI =!= True || MY.Has_MPI =?= True) && \
-        ($(START_EXTRA))
+        (START_EXTRA ?: true) && (OSG_PROJECT_RESTRICTION ?: true)
```
matyasselmeci, line 68: Where do you advertise START_EXTRA? If START_EXTRA is not an advertised attribute then you need the `$( )` around it to do the macro expansion.

rynge: `OSG_PROJECT_RESTRICTION` must have the same problem then. Would it be enough to add them to `STARTD_ATTRS`?

rynge: Adding both to `STARTD_ATTRS` fixed it. Tested and added to this PR.

matyasselmeci (approved):
LGTM.

PR #127, opened 2023-02-23
Install the new job hook

matyasselmeci:
Off by default unless CONTAINER_PILOT_USE_JOB_HOOK is set

Depends on https://github.com/opensciencegrid/osg-flock/pull/220 being merged first


rynge (approved):
LGTM

PR #129, opened 2023-03-08
Turn on the prepare-job hook by default

matyasselmeci:
I added the attribute OSG_USING_JOB_HOOK in osg-flock so we can track where this is deployed.


rynge (approved):
LGTM

PR #130, opened 2023-03-13
Use osdf-client from RPMs instead of downloading our own from the osg-flock repo (OSPOOL-64)

matyasselmeci:
Use the stash/osdf plugin 'shipped with' Condor instead of having a copy in the osg-flock repo. For the backfill containers this is the RPM from the OSG repos; for glideins, the condor tarball will contain it.

We still test it, it's just that instead of enabling it when the tests pass, we disable it when the tests fail.

This is temporary - I'll move the logic into the `additional-htcondor-config` script later.

https://opensciencegrid.atlassian.net/browse/OSPOOL-64


matyasselmeci:
Good call. Fixed.

PR #132, opened 2023-03-20
Environment tidying

matyasselmeci:
Tidy up the environment variables:
- Move all of the ENV statements to the bottom of the Dockerfile to keep them together
- Move some of the defaults from 10-setup-htcondor.sh to the Dockerfile
- Use consistent booleans


rynge (approved):
Nice cleanup!

PR #134, opened 2023-03-29
Add the environment variable `SINGULARITY_BIND_EXTRA` (SOFTWARE-5528)

matyasselmeci:
This provides a way of adding additional bind-mounts into Singularity jobs.  This is something we can already do in factory pilots with `GLIDEIN_SINGULARITY_BINDPATH`.

`SINGULARITY_BIND_EXTRA` adds the bind-mounts by setting `GLIDEIN_SINGULARITY_BINDPATH` in the glidein config; I didn't want to couple the behavior to the implementation which is why I gave it a different name.


rynge (approved):
LGTM

PR #135, opened 2023-05-02
GLIDEIN_ToRetire and GLIDEIN_ToDie

rynge:
This adds two new attributes: `GLIDEIN_ToRetire` and `GLIDEIN_ToDie`. We need these to align with the GWMS glideins in the pool. The attributes will not only be used within the container, but also for monitoring (OSPOOL-72 for example) and potentially for scheduling decisions (when we implement better logic for `JobDurationCategory`).

### DIFF
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -134,6 +134,24 @@ if [ "x$GLIDEIN_Start_Extra" != "x" ]; then
 else
     export GLIDEIN_Start_Extra="True"
 fi
+if ! [[ "$ACCEPT_JOBS_FOR_HOURS" =~ ^[0-9]+$ ]]; then
+    echo "ACCEPT_JOBS_FOR_HOURS has to be a positive integer" 1>&2
+    exit 1
+else
+    if [ $ACCEPT_JOBS_FOR_HOURS -le 0 ]; then
+        echo "ACCEPT_JOBS_FOR_HOURS has to be a positive integer" 1>&2
+        exit 1
+    fi
+fi
+if ! [[ "$RETIREMENT_HOURS" =~ ^[0-9]+$ ]]; then
+    echo "RETIREMENT_HOURS has to be a positive integer" 1>&2
+    exit 1
+else
+    if [ $RETIREMENT_HOURS -le 0 ]; then
+        echo "RETIREMENT_HOURS has to be a positive integer" 1>&2
+        exit 1
+    fi
+fi
 
 
 #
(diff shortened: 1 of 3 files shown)


10-setup-htcondor.sh
```diff
@@ -367,6 +367,13 @@ if is_true "$CONTAINER_PILOT_USE_JOB_HOOK" && [[ ! -e ${prepare_hook} ]]; then
     exit 1
 fi
 
+# use GLIDEIN_ToRetire - this is for aligning with GWMS glideins
+NOW=$(date +'%s')
+GLIDEIN_ToRetire=$(($NOW + $ACCEPT_JOBS_FOR_HOURS * 60 * 60))
+
+# Give the instance 24 hours to finish up before exiting
+GLIDEIN_ToDie=$(($GLIDEIN_ToRetire + 24 * 60 * 60))
```
dsschult, line 375: Could you make the "24 hours" a variable?  I have sites that have a hard time limit, so would like to guarantee the pilot dies within a certain amount of time.

Not sure if I prefer having to add two variables to get the total time limit, or if it would be nicer to define the ToDie time directly.

rynge: Good point! I added a new env var named `RETIREMENT_HOURS`.

Dockerfile
```diff
@@ -187,6 +187,8 @@ ENV CVMFS_QUOTA_LIMIT=
 
 # How many hours to accept new jobs for
 ENV ACCEPT_JOBS_FOR_HOURS=336
+# Hours to let running jobs finish during retirement
+env RETIREMENT_HOURS=24
```
matyasselmeci, line 191:
```suggestion
ENV RETIREMENT_HOURS=24
```

10-setup-htcondor.sh
```diff
@@ -367,6 +367,20 @@ if is_true "$CONTAINER_PILOT_USE_JOB_HOOK" && [[ ! -e ${prepare_hook} ]]; then
     exit 1
 fi
 
+cat <<EOF
```
matyasselmeci, line 370: Can you add some validation? Bail out if $ACCEPT_JOBS_FOR_HOURS or $RETIREMENT_HOURS are not positive integers?

matyasselmeci (approved):
A few suggestions but otherwise pre-approving. BTW, I like the `cat <<EOF` help text.

10-setup-htcondor.sh
rynge: Added

PR #136, opened 2023-05-08
More debugging for the Docker workflow

matyasselmeci:


brianhlin:
Unless I'm missing something, this appears to leave the container running between tests. It may make sense to make sure that we don't run subsequent tests in a new container to prevent pollution from previous tests.

brianhlin (approved):
Can't get any more broken, I guess? Seems to do what it says on the tin

PR #140, opened 2023-05-12
Add option to use the highest versioned tag of CVMFSEXEC

matyasselmeci:
Caveat: this adds the requirement for git >= 2 for the `--sort=version:refname` feature, which is not available on el7.


rynge (approved):
LGTM

PR #143, opened 2023-07-10
SOFTWARE-5538: update github actions to build el9 image

mwestphall:


.github/workflows/build-container.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           OS: ${{ matrix.os }}
         run: |
           PREFIX="output_image=${GITHUB_REPOSITORY}:${SERIES}"
-          if [[ $OS == 'el8' ]]; then
+          if [[ $OS == 'el9' ]]; then
```
mwestphall, line 28: This change should change the "no os" tag from el8 to el9. Is that the desired behavior or should it be kept on el8?

brianhlin:
```suggestion
          if [[ $OS == 'el8' ]]; then
```

Hrm yeah, I don't think we can pull the trigger on this OS update just yet.

brianhlin (changes requested):
One in-line change requested

Could you also update the PR title + commit to include the Jira ticket key?

brianhlin (approved):
LGTM

PR #145, opened 2023-07-26
Advertise to both dev PATh collectors

matyasselmeci:
The second one is on river.


matyasselmeci:
tests using the cuda image failed due to running out of disk which shouldn't be affected by this commit.

PR #147, opened 2023-08-03
Skip singularity test for cuda images

matyasselmeci:
Running with singularity requires converting the image to SIF first and the runner doesn't have the disk space to do that.

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -72,9 +72,14 @@ jobs:
         env:
           CONTAINER_IMAGE: ${{ steps.build-image.outputs.timestamp-image }}
         run: |
-          sudo ./tests/test_inside_gha.sh singularity \
-                                          bindmount \
-                                          "$CONTAINER_IMAGE"
+          if [[ $CONTAINER_IMAGE == *cuda* ]]; then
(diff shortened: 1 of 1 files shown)


dsschult:
We have that issue with some of our actions too, and while there are a couple workarounds to get more disk space on github runners, the long-term solution is to do self-hosted runners.  If you want to continue these tests, I assume there is hardware somewhere in PATh for this?

matyasselmeci:
Did you set up dedicated machines for that?  We were thinking of some sort of job or pilot...

dsschult:
> Did you set up dedicated machines for that? We were thinking of some sort of job or pilot...

Yeah, the easiest is a VM, though you could do it in k8s.  The github runner comes with a systemd service, thus why a VM is easy.

A job might be hard, since you want the runner to be available all the time to receive new work.  But I guess if you don't mind some interruptions, you could put it in a job that runs forever.  I just think I'd get frustrated if the tests aren't running because the job got evicted, and I need to wait half an hour for it to restart.

matyasselmeci:
We'll give that a shot at some point but until then we should just disable this test.

brianhlin (approved):
L"G"TM

PR #148, opened 2023-08-08
Change the default image OS from el8 (i.e., CentOS Stream 8) to al8 (i.e., Alma 8)

matyasselmeci:
This changes the OS-less tags (such as `3.6-release`) to be equivalent to the corresponding al8 tags (so `3.6-al8-release`).  It keeps the `al8` tags for backward compat, in case anybody is using those, it just adds the OS-less tags as aliases.

Note that this has the side effect of making tags like `3.6-el8-release` for the CentOS Stream 8-based builds, which we didn't have before.

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -31,11 +31,7 @@ jobs:
           OS: ${{ matrix.os }}
         run: |
           PREFIX="output_image=${GITHUB_REPOSITORY}:${SERIES}"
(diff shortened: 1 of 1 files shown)


matyasselmeci:
Tested alma 8 on a pod in tiger (TIGER-OSG-BACKFILL-PROD)

brianhlin (approved):
LGTM

PR #149, opened 2023-09-06
Install socat into the image for network testing

matyasselmeci:
socat is useful for testing primitive network connectivity in the EP and is only 1 MB.


rynge (approved):
LGTM

PR #151, opened 2023-09-26
Add OSG 23 images (SOFTWARE-5763)

brianhlin:
Also drop specific `el8` tags


mwestphall (approved):
LGTM

brianhlin:
I've removed the 3.6 EL8 status check that was required since we no longer build those images.

PR #152, opened 2023-10-02
un-exclude osg 23 testing and release repos from GHAs 

mwestphall:


matyasselmeci:
It looks like after I resolved the merge conflict, this PR had no differences from master, so I think the changes already made it in. I'm going to close it.

PR #156, opened 2023-10-20
If POOL is a hostname or host:port then use it as a central manager as opposed to bailing out

matyasselmeci:
We don't specify a separate CCB_ADDRESS in that case; also we disable remote syslog. Also toss in a special case that if POOL is "ospool" then we choose prod-ospool or itb-ospool depending on $ITB.  I like this better than having that behavior be implicit if POOL is blank.

(SOFTWARE-5383)


rynge (approved):
LGTM

dsschult:
I'm in support.  This makes the IceCube config easier.

PR #157, opened 2023-10-20
Better check for making sure no CVMFSEXEC_REPOS were specified

matyasselmeci:
Make sure there is at least one letter in $CVMFSEXEC_REPOS so we don't think a repo is specified for a string like ` ` or `,` which could occur if $CVMFSEXEC_REPOS is built up by a script, for example.


brianhlin (approved):
LGTM

PR #158, opened 2023-10-20
Allow adding arbitrary extra STARTD attributes (SOFTWARE-5324)

matyasselmeci:
If you bind-mount a file containing `key=value` pairs to `/etc/osg/extra-attributes.cfg` then those attributes will be added to STARTD_ATTRS.  A new script, `/usr/local/sbin/add-extra-attributes`, parses the file and does some light validation, then appends the attributes to the pilot config file.

I put the script in a new sbin/ directory so it doesn't get mixed in with all the other random files we have; I want to reorganize the repo in a later PR to put the various config and script files in directories as opposed to having them all be at the top level.

### DIFF
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -651,7 +651,7 @@ use POLICY : WANT_HOLD_IF( DISK_EXCEEDED, \$(HOLD_SUBCODE_DISK_EXCEEDED:104), \$
 END
 fi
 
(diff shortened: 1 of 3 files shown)


rynge (approved):
LGTM. This will be useful!

First I was wondering why we are not just doing a HTCondor config directly, but I then realized that the /pilot directory approach prevents that.

PR #159, opened 2023-11-09
Reorganize binaries

matyasselmeci:
Reorganize the repo by moving the various scripts that we install to `sbin/` which we copy in to `/usr/local/sbin/` inside the container instead of copying the scripts in one at a time.

This changes where some of the files are located inside the container, so various references needed to be adjusted too.


rynge (approved):
LGTM

PR #160, opened 2023-11-16
Use tini when using cvmfsexec to handle signal propagation

matyasselmeci:
We have an issue where if the container uses cvmfsexec to mount repos, then a SIGTERM (as sent by docker stop) won't shut it down and we have to wait for docker to time out and send SIGKILL.

This is because cvmfsexec is pid 1 inside the container, which is special, and we have to have it explicitly forward the signals to the child processes.  If we don't use cvmfsexec, then supervisord is pid 1, and signal propagation is handled.

This commit adds `tini`, a small init system, to handle signal propagation; it is only used when we're using cvmfsexec.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,7 @@ RUN useradd osg \
         rsyslog rsyslog-gnutls python3-cryptography python3-requests \
         bind-utils \
         socat \
+        tini \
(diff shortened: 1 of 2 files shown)


rynge (approved):
Nice fix! It is a little bit odd to have both tini and supervisord in the same container, but whatever works.

PR #161, opened 2023-12-13
Replaced CPUJOB_ON_GPUSLOT with the new RoomForCPUOnlyJobs

rynge:
The OSPool GPU glideins have switched to an expression to help backfill partionable GPU slots with CPU-only jobs, and leave some resources for potential future GPU jobs. `RoomForCPUOnlyJobs` is defined in:

https://github.com/opensciencegrid/osg-flock/blob/master/ospool-pilot/main/pilot/additional-htcondor-config#L95-L99

This PR aligns osgvo-docker-pilot with those glideins. The `ALLOW_CPUJOB_ON_GPUSLOT` configuration flag is removed as it never worked well.

### DIFF
--- a/10-setup-htcondor.sh
+++ b/10-setup-htcondor.sh
@@ -501,14 +501,6 @@ if [ "x$OSG_SQUID_LOCATION" != "x" ]; then
     export http_proxy="$OSG_SQUID_LOCATION"
 fi
 
-# some admins prefer to reserve gpu slots for gpu jobs, others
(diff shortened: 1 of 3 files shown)


matyasselmeci (approved):
The START expression change doesn't quite match what's in our frontend config; see suggestion. I'm pre-approving though.

rynge:
Thanks - start expression is slightly different because the FE group is for only GPU resources, while the container has be able to run well on both non-GPU and GPU resources.

PR #163, opened 2024-02-01
Exit out if 'exec' fails

matyasselmeci:
I ran into two issues while trying to use this image with `apptainer run`:
- `apptainer run` put `--` into the first argument, causing `exec "$@"` to fail because it wasn't a real option
- the exec command failing did _not_ cause the script to stop


matyasselmeci:
OK, ready for review. Turns out apptainer wasn't doing anything weird, it's just I was doing `apptainer run image.sif --contain ...` instead of `apptainer run --contain image.sif ...` and I was getting what I asked for.  We should still handle exec failing.

brianhlin (approved):
LGTM

PR #165, opened 2024-02-28
Add knob to disable adding the expanse user and group (u532362:g12497)

matyasselmeci:
This user has such a high UID that on some systems pulling or running with podman fails.  Adding the user is still the default (because I can't add arbitrary build args in the GHA).


brianhlin (approved):
LGTM

PR #167, opened 2024-04-25
Fix error message when failing to set mandatory variable

matyasselmeci:
The variable $error_gen does not actually exist; it's supposed to be the path to the error_gen.sh script but that doesn't print anything in a pilot container. Just print the error message ourselves.


rynge (approved):
LGTM

PR #171, opened 2024-07-05
Fix failure to detect HAS_SINGULARITY in images based on Condor 23.8+…

matyasselmeci:
… based on `condor_status -direct` issues

Condor 23.8 has a bug where condor_status -direct for startd ads still attempts to contact the collector.  Hopefully it will be fixed in 23.10. For those versions, use a hack where we use -pool instead of -direct.


matyasselmeci:
Merged this so we get some builds going.

PR #162, opened 2024-01-16
Re-enable OSG 23 development images (SOFTWARE-5673)

brianhlin:

PR #172, opened 2024-07-10
Bump condor max version requirement for -pool workaround to include 23.10.0 dailies

matyasselmeci:


matyasselmeci:
Should fix #162

brianhlin (approved):
LGTM

PR #173, opened 2024-07-11
Drop 3.6 and al8 from build matrix

matyasselmeci:
Drop OSG 3.6 (EOL) and al8 OS (no longer built) from the build matrix.


rynge (approved):
Great!

PR #175, opened 2024-09-09
Some tweaks to add-extra-attributes script:

matyasselmeci:
- use start kwarg in enumerate() to avoid having to add 1 to the index each time
- allow spaces around the `=` between key and value


rynge (approved):
LGTM

PR #182, opened 2025-01-03
Use an osdf:// URL for the remote file for OSDF validation

matyasselmeci:
Current plugin version (7.11.7) fails if the federation isn't specified and can't be determined from the URL.


rynge (approved):
LGTM

PR #183, opened 2025-02-19
Delete quotes in/around CVMFSEXEC_REPOS

matyasselmeci:
People passing in env vars via a docker --env-file may be surprised to learn that quotes are passed through literally.  Handle that case by just deleting the quotes.


rynge (approved):
LGTM

PR #187, opened 2025-04-23
Drop redundant OSDF plugin detection

matyasselmeci:
Apparently, the scripts in osg-flock have taken care of it since version 741 or so: https://github.com/opensciencegrid/osg-flock/pull/272


rynge (approved):
LGTM

PR #188, opened 2025-04-25
SOFTWARE-6138: Write apptainer image cache config on startup

mwestphall:


20-setup-apptainer.sh
```diff
@@ -0,0 +1,21 @@
+# Configure apptainer based on environment variables
+# Currently, just set the registry mirror
+# Note that this only works in apptainer >=1.3, <1.4
+
+set -xe
```
matyasselmeci, line 5: Careful! These scripts are sourced, not run, so any shell options set by one script will be  kept for subsequent scripts.

PR #190, opened 2025-05-28
OSPOOL-151: Enable garbage collection for ep containers

rynge:
This is applying the same cleanup tool to containers, as we use in the glideins. For now, this is an opt-in functionality which can be enabled by passing `GARBAGE_COLLECTION=1` into the container environment.


matyasselmeci (approved):
LGTM

PR #192, opened 2025-06-12
OSPOOL-151: Change garbage collection to be on by default (opt-out)

rynge:


brianhlin (approved):
This LGTM but I'm not sure why we've got so many failing tests...

EDIT: Outside of the CUDA images, all the failing tests were amd64. The successful tests were arm64.

rynge:
I wonder if the workflow is unable to load the token for PRs?

brianhlin:
Nope, we don't have an IDTOKEN secret and don't need to report to a CM for these tests. It looks like we've set up a dummy token since the very beginning https://github.com/opensciencegrid/osgvo-docker-pilot/blob/master/tests/test_inside_gha.sh#L4.

This does feel like an HTCondor behavior change to make advertising failures fatal on the EP side? I also was digging into versions and had this written up:

> It looks like we last saw across the board successes with `24.9.0-0.806635.el9` ([test run](https://github.com/opensciencegrid/osgvo-docker-pilot/actions/runs/14997588611/job/42135595963#step:4:1133)) and started seeing these failures with `24.9.0-0.809873.el9` ([test run](https://github.com/opensciencegrid/osgvo-docker-pilot/actions/runs/15516287897/job/43683592741#step:4:1158)) and `23.10.27-0.811708.el9` ([test run](https://github.com/opensciencegrid/osgvo-docker-pilot/actions/runs/15593284271/job/43923865593#step:4:1157))

But then saw that we're seeing failures on `23.10.25-1.el9` ([test run](https://github.com/opensciencegrid/osgvo-docker-pilot/actions/runs/15593284271/job/43923865628#step:4:1147) but successes from the same version from 2 days ago ([test run](https://github.com/opensciencegrid/osgvo-docker-pilot/actions/runs/15571032381/job/43846691683#step:4:1172)). Were there any CM changes between today and 2d ago?

It looks like we just skip the tests for the ARM hosts so that explains the arch-based failure pattern.

matyasselmeci:
We've been seeing these errors (but only with 24-development) since May 25 (example: https://github.com/opensciencegrid/osgvo-docker-pilot/actions/runs/15235657588) but the new osg-ca-certs RPMs were only built on 6/9 and 6/10.

brianhlin:
I've got half a fix here https://github.com/opensciencegrid/osgvo-docker-pilot/compare/master...brianhlin:osgvo-docker-pilot:troubleshoot-failing-ci. It appears that our startd-ready-detection isn't working (or the startd is taking longer to startup than before). Passed off the finishing touches to be done by @mwestphall . Once that's been merged into `master`, we'll want Mats to rebase.

PR #193, opened 2025-06-18
Handle SIGTERM and SIGINT with condor_off, allow for off strategies

rynge:
`SIGTERM` and `SIGINT` will be trapped and mapped to the appropriate `condor_off` command.

You can choose strategy by setting the `SHUTDOWN_STRATEGY` environment variable to one of `peaceful`, `graceful` or `fast`. 

`graceful` provides 24 hours for jobs to finish.

`fast` is default for now, matching the current behavior. We could have policy discussion on how for example the RPM should shutdown. For K8s deployments, `graceful` with something like this works well:

    spec:
      terminationGracePeriodSeconds: 86400
      containers:
      ...
        env:
          - name: SHUTDOWN_STRATEGY
            value: "graceful"

### DIFF
--- a/10-htcondor.conf
+++ b/10-htcondor.conf
@@ -1,7 +1,12 @@
 [program:condor_master]
 command=/usr/local/sbin/condor_master_wrapper
-autorestart=unexpected
+autorestart=false
 startsecs=60
 stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
 redirect_stderr=true
+
+; wait for condor_off peaceful/graceful to finish
+stopsignal=TERM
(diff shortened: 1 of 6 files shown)


matyasselmeci:
This looks cool! It's the school so I'm not going to merge anything, but I'd like to try the branch out in ITB.

rynge:
I changed the fast shutdown to target the master daemon, and that get the almost instant kill. For the others, I think it is correct to target startd and let the master shutdown ones everything else has settled.

The delay comes from the `DAEMON_SHUTDOWN_FAST` which only gets evaluated when there is an ad update.

PR #196, opened 2025-06-23
INF-2962: install Pelican client with debug symbols

brianhlin:
We don't know where we may need this a priori and pelican-debug is
only an additional 20MB at the time of this commit

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,11 +44,15 @@ RUN useradd osg \
 # Pull HTCondor from the proper repo. For "release" we need to use
 # osg-upcoming-testing to meet the patch tuesday requirements.
 RUN if [[ $BASE_YUM_REPO = release ]]; then \
-      yum -y --enablerepo=osg-upcoming-testing swap pelican pelican-debug && \
-      yum -y --enablerepo=osg-upcoming-testing install condor; \
+      yum -y --enablerepo=osg-upcoming-testing install condor && \
+      if /usr/local/bin/pkg-cmp-gt.sh condor 23.10.26; then \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -44,9 +44,11 @@ RUN useradd osg \
 # Pull HTCondor from the proper repo. For "release" we need to use
 # osg-upcoming-testing to meet the patch tuesday requirements.
 RUN if [[ $BASE_YUM_REPO = release ]]; then \
-      yum -y --enablerepo=osg-upcoming-testing install condor; \
+      yum -y --enablerepo=osg-upcoming-testing install condor \
+                                                       pelican-debug; \
```
matyasselmeci, line 48:
```suggestion
      yum -y --enablerepo=osg-upcoming-testing swap pelican pelican-debug && \
      yum -y --enablerepo=osg-upcoming-testing install condor; \
```
Apparently 'swap' works even if the original package is not installed.

Dockerfile
```diff
@@ -44,9 +44,11 @@ RUN useradd osg \
...
     else \
-      yum -y install condor; \
+      yum -y install condor \
+                     pelican-debug;\
```
matyasselmeci, line 51:
```suggestion
      yum -y swap pelican pelican-debug && \
      yum -y install condor; \
```

matyasselmeci (commented):
See if my suggestions fix the errors

matyasselmeci:
Looks like they failed on 23 release and testing. One sec...

matyasselmeci (approved):

PR #201, opened 2025-08-12
Tests: factor out startd_addr waiting/detection

rynge:
- startd_addr waiting/detection is now a reusable script baked into the container
- This solves some waiting corner cases, such as making sure we are not querying the master before it has even started up. I also added a wait for the startd log, which takes care of the gap between master and startd starting up.
- Having it as script inside just makes for cleaner code in the tests.

### DIFF
--- a/sbin/startd_addr.sh
+++ b/sbin/startd_addr.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+CONDOR_LOGDIR=/pilot/log
+
+function condor_version_in_range {
+    local minimum maximum
+    minimum=${1:?minimum not provided to condor_version_in_range}
+    maximum=${2:-99.99.99}
+
+    local condor_version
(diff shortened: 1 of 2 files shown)


sbin/startd_addr.sh
```diff
@@ -0,0 +1,61 @@
...
+    local condor_version
+    condor_version=$(condor_version | awk '/CondorVersion/ {print $2}')
+    python3 -c '
+import sys
+minimum = [int(x) for x in sys.argv[1].split(".")]
+maximum = [int(x) for x in sys.argv[2].split(".")]
+version = [int(x) for x in sys.argv[3].split(".")]
+sys.exit(0 if minimum <= version <= maximum else 1)
+' "$minimum" "$maximum" "$condor_version"
+}
```
brianhlin, line 19: It doesn't look like this is used at all?

brianhlin (approved):
Thanks for the PR! One minor comment but otherwise LGTM.

PR #213, opened 2026-02-05
Use a smaller default image inside the CI

matyasselmeci:
This avoids timeouts caused by slow pulls of a large image.


brianhlin (approved):
LGTM

PR #215, opened 2026-02-12
Remove compat config; no longer necessary

matyasselmeci:
This code path doesn't get traversed because we put ${osgvo_additional_htcondor_config} into the image, so it's just dead code.


rynge (approved):
Looks good!

